PR #2, opened 2017-08-25
Use Grok instead of OpenJpeg

whikloj:
Partially Resolves? https://github.com/Islandora-CLAW/CLAW/issues/703

This removes the OpenJpeg library and instead downloads and compiles the [Grok fork](https://github.com/GrokImageCompression/grok) for use with Cantaloupe.


whikloj:
Oops forgot instructions.

1. Checkout this branch
1. `vagrant up --no-provision`
1. `ansible-playbook -i inventory/vagrant bootstrap.yml`
1. `ansible-playbook -i inventory/vagrant playbook.yml --tags cantaloupe,tomcat8,grok`
1. `vagrant ssh`
1. `sudo mkdir -p /var/www/html/drupal/web/`
1. Put test images in /var/www/html/drupal/web (tiff, jp2, etc)
1. Check that you can see them at: http://localhost:8080/cantaloupe/iiif/2/FILE.NAME/full/full/0/default.jpg
1. Check logs at: /var/log/cantaloupe

inventory/vagrant/group_vars/all.yml
```diff
@@ -10,3 +10,7 @@ islandora_extra_packages:
   - unzip
   - build-essential
   - vim
+  - cmake
```
jonathangreen: Since these packages are specific to the grok role. I would install them inside the grok role. 

Maybe inside install.yml using a structure like this: 
https://github.com/Islandora-Devops/claw-playbook/blob/66ef19d323a3c9cc1e1d3e88438d99d3abb0d6e1/roles/internal/openjpeg/tasks/install.yml#L3-L8

roles/internal/grok/files/etc/ld.so.conf.d/libgrok.conf
```diff
@@ -0,0 +1 @@
+/usr/local/lib
```
jonathangreen: Is this actually needed? I believe `/usr/local/lib` is in the library path by default.

whikloj:
And it works with @dannylamb's terrible Tokyo JP2 from https://github.com/Islandora-CLAW/CLAW/issues/572#issuecomment-310434575

roles/internal/grok/tasks/install.yml
```diff
@@ -0,0 +1,36 @@
...
+    chdir: "{{ grok_clone_directory }}/build"
+
+- name: Install grok
+  make:
+    chdir: "{{ grok_clone_directory }}/build"
+    target: install
+
+- name: Add Grok libraries to ld.so.conf.d
+  copy: src={{ item }} dest=/{{ item }} owner=root group=root mode=0444
+  with_items:
```
jonathangreen: See comment about this not being necessary above. If we do include it, I wouldn't use a loop to install one item. 

I'd rather see something like: 
```yaml
- name: Add Grok libraries to ld.so.conf.d
  copy: 
    src: libgrok.conf 
    dest: /etc/ld.so.conf.d/libgrok.conf 
    owner: root 
    group: root 
    mode: "0444"
```

or alternatively in the single line syntax
```yaml
- name: Add Grok libraries to ld.so.conf.d
  copy: src=libgrok.conf dest=/etc/ld.so.conf.d/libgrok.conf owner=root group=root mode=0444
```

I personally prefer the YAML syntax to the single line, but thats just a style thing.

roles/internal/grok/vars/main.yml
```diff
@@ -0,0 +1,4 @@
+grok_clone_directory: /opt/grok
+grok_version_tag: v2.3.0
```
jonathangreen: I would put both of these in the `roles/internal/grok/defaults/main.yml` file, as they are super hard to override where they are. Here is a [good reference](https://codereviewvideos.com/course/ansible-tutorial/video/variable-precedence-where-to-put-your-role-vars) talking about where to put variables. The TL;DR is that you should always use the defaults, not vars.

roles/internal/grok/vars/main.yml
```diff
@@ -0,0 +1,4 @@
...
+
+cantaloupe_OpenJpegProcessor_path_to_binaries: /usr/local/bin
```
jonathangreen: This shouldn't be defined in this playbook for modularity. You want this role to stand on its own not to be setting variables from other roles. I would set this in this with the rest of the cantaloupe variables in: 
https://github.com/Islandora-Devops/claw-playbook/blob/master/inventory/vagrant/group_vars/tomcat.yml#L37-L53

roles/internal/grok/tasks/install.yml
```diff
@@ -0,0 +1,36 @@
...
+  cmake:
+    binary_dir: "{{ grok_clone_directory }}/build"
+    source_dir: "{{ grok_clone_directory }}"
+    build_type: Release
+      
+- name: Make grok
+  make:
+    chdir: "{{ grok_clone_directory }}/build"
+
+- name: Install grok
```
jonathangreen: Did you run the role twice to make sure this is idempotent? Taking a quick glance at the cmake library that is included, I don't think that this will be idempotent without the creates statement, but I could be reading the code wrong and I haven't actually run it yet.

You can tell if its idempotent by running it twice and seeing if ansible reports any tasks as changed.

jonathangreen (changes requested):
Just a few little nits the with the playbook, but it mostly looks good.

roles/internal/grok/files/etc/ld.so.conf.d/libgrok.conf
whikloj: I seemed to need it to decode JP2s, but Tiffs worked without it. Perhaps I just need the to run `ldconfig`?

roles/internal/grok/tasks/install.yml
whikloj: So it does seem to have 'changed' it. But I also realized reading the code that he has a _creates_ option, so I'll try adding that.

roles/internal/grok/vars/main.yml
whikloj: Right, I did have these there and then I stupidly moved the entire file to cause the below Cantaloupe option to get set. I'll fix them both.

jonathangreen:
One other minor thing I forgot to mention is that I try to avoid python libraries as much as possible, simply because of the extra cognitive overhead of having to deal with python as well as all the ansible yaml, unless the library buys something thats hard to do in YAML. The big on is if a library provides idempotency, its probably worth it.

In this case I'm not sure it buys us idempotency and I'm not sure its much more convenient  then just doing something like: 
```yaml
- name: Cmake grok
  command:
    cmake blah blah
  args:
    creates: file blah blah 
```

Just something else to consider.

roles/internal/grok/files/etc/ld.so.conf.d/libgrok.conf
jonathangreen: For sure will need `ldconfig` to load the libraries.

I think it should be included by default, I guess its not a big deal to define it twice.

whikloj:
@jonathangreen ok I think I got it all. I do need to run `ldconfig` at the end or the new libraries are not picked up. This makes that process have one always _changed_ task. But otherwise it seems to be idempotent now.

whikloj:
Forgot some unused/unneeded variables and files.

roles/internal/grok/tasks/install.yml
```diff
@@ -0,0 +1,43 @@
...
+    chdir: "{{ grok_clone_directory }}/build"
+    creates: "{{ grok_clone_directory }}/build/bin/opj_decompress"
+
+- name: Install grok
+  command: make install
+  args:
+    chdir: "{{ grok_clone_directory }}/build"
+    creates: /usr/local/bin/opj_decompress
+
+- name: Update ldconfig
```
jonathangreen: So to get rid of the last idempotency thing, I'd do something like this:

```yaml
- name: Install grok
  command: make install
  args:
    chdir: "{{ grok_clone_directory }}/build"
    creates: /usr/local/bin/opj_decompress
    register: grok_make_install

- name: Update ldconfig
  command: ldconfig
  when: grok_make_install.changed is defined AND grok_make_install.changed
```

roles/internal/grok/tasks/install.yml
```diff
@@ -0,0 +1,43 @@
...
+  git:
+    repo: https://github.com/GrokImageCompression/grok.git
+    dest: "{{ grok_clone_directory }}"
+    version: "{{ grok_version_tag }}"
+
+- name: Make build directory
+  file:
+    path: "{{ grok_clone_directory }}/build"
+    state: directory
+    mode: 0755
```
jonathangreen: weird ansible gotcha here, but always put the mode in quotes since its an octal number or it'll be interpreted as decimal, and shit will get weird haha.

https://github.com/ansible/ansible/issues/9196
https://blog.tyk.nu/blog/ansible-and-unix-file-permissions/

jonathangreen (commented):
Looks good man. Couple other small things, but after that it should be good to go.

roles/internal/grok/tasks/install.yml
whikloj: Your `when:` line above caused an error when I tried to run it. I was able (I think) to get the expected process with `when: grok_make_install.changed`. Does that seem viable?

jonathangreen (approved):

PR #5, opened 2017-08-29
Switching up how we do composer install and adding carapace.

dannylamb:
Resolves Islandora-CLAW/CLAW#700

### Testing
Vagrant up and check that carapace is installed and the default theme at http://localhost:8000


jonathangreen:
vagrant up!

jonathangreen:
Looks good to me. Came up clean and theme was enabled.

PR #6, opened 2017-08-29
activemq

Natkeeran:
## Ticket 
https://github.com/Islandora-CLAW/CLAW/issues/690

## What it does
Adds a role to install activemq.  Implements the  [activemq.sh](https://github.com/Islandora-CLAW/claw_vagrant/blob/master/scripts/activemq.sh) logic in Ansible.

## Testing
* Get PR and vagrant up
* Go to http://localhost:8161/admin (admin:admin), verify Activemq is up and running

If you already have a claw-playbook provisioned: 
```
vagrant up --no-provision
ansible-playbook -i inventory/vagrant/hosts  playbook.yml --tags activemq
```

### DIFF
--- a/roles/internal/activemq/defaults/main.yml
+++ b/roles/internal/activemq/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+
+activemq_download_dir: /usr/local/src
+activemq_version: 5.14.5
+activemq_install_root: /opt
+activemq_install_symlink: /opt/activemq
+activemq_user: ubuntu
+activemq_group: ubuntu
+
--- a/roles/internal/activemq/defaults/main.yml
+++ b/roles/internal/activemq/defaults/main.yml
@@ -1,9 +1,7 @@
 ---
 
-activemq_download_dir: /usr/local/src
 activemq_version: 5.14.5
 activemq_install_root: /opt
-activemq_install_symlink: /opt/activemq
 activemq_user: ubuntu
 activemq_group: ubuntu
 
--- a/roles/internal/activemq/tasks/install.yml
+++ b/roles/internal/activemq/tasks/install.yml
@@ -1,8 +1,4 @@
 ---
-- name: Clean activemq path
-  file:
-    state: absent
-    path: "{{ activemq_install_root }}/activemq"
 
 - name: Download and unarchive activemq
   unarchive:
@@ -11,18 +7,18 @@
     dest: "{{ activemq_install_root }}"
     owner: "{{ activemq_user }}"
     group: "{{ activemq_group }}"
+    creates: "{{ activemq_install_root }}/activemq"
 
 - name: Move activemq
   command: mv "{{ activemq_install_root }}/apache-activemq-{{ activemq_version }}" "{{ activemq_install_root }}/activemq"
+  args:
+    creates: "{{ activemq_install_root }}/activemq"
 
 - name: Create symlink
   file:
     state: link
     src: "/opt/activemq/bin/activemq"
     dest: "/etc/init.d/activemq"
-
-- name: Add activemq service to start up
-  shell: 'update-rc.d activemq defaults'
 
 - name: Start activeMQ service
   service:
--- a/roles/internal/activemq/tasks/install.yml
+++ b/roles/internal/activemq/tasks/install.yml
@@ -1,6 +1,6 @@
 ---
 
-- name: Download and unarchive activemq
+- name: Download and unarchive
   unarchive:
     remote_src: yes
     src: http://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz
@@ -20,8 +20,9 @@
     src: "/opt/activemq/bin/activemq"
     dest: "/etc/init.d/activemq"
 
-- name: Start activeMQ service
-  service:
+- name: Reload systemd
+  systemd:
+    daemon-reload: yes
+    enabled: yes
     state: started
     name: activemq
-    enabled: yes
--- a/roles/internal/activemq/tasks/install.yml
+++ b/roles/internal/activemq/tasks/install.yml
@@ -1,0 +1,31 @@
+---
+- name: Clean activemq path
+  file:
+    state: absent
+    path: "{{ activemq_install_root }}/activemq"
+
+- name: Download and unarchive activemq
+  unarchive:
+    remote_src: yes
+    src: http://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz
+    dest: "{{ activemq_install_root }}"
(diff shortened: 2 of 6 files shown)


roles/internal/activemq/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+activemq_download_dir: /usr/local/src
+activemq_version: 5.14.5
+activemq_install_root: /opt
+activemq_install_symlink: /opt/activemq
```
jonathangreen: This variable seems to be unused.

roles/internal/activemq/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+activemq_download_dir: /usr/local/src
```
jonathangreen: This seems to be unused.

roles/internal/activemq/tasks/install.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Clean activemq path
```
jonathangreen: What is this task meant to do? 

Right now it seems like every time the playbook is run its deleting activemq and starting again. We don't want to do that. We want the playbook to be idempotent, so every time its run it won't make changes if the task has already been run. Perhaps this task can just be removed?

roles/internal/activemq/tasks/install.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Download and unarchive activemq
+  unarchive:
+    remote_src: yes
+    src: http://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz
+    dest: "{{ activemq_install_root }}"
+    owner: "{{ activemq_user }}"
+    group: "{{ activemq_group }}"
+
+- name: Move activemq
+  command: mv "{{ activemq_install_root }}/apache-activemq-{{ activemq_version }}" "{{ activemq_install_root }}/activemq"
```
jonathangreen: This needs to be made idempotent somehow, perhaps using the when or creates statement statement.

Perhaps something like this: 
```yaml
- name: Move activemq
  command: mv "{{ activemq_install_root }}/apache-activemq-{{ activemq_version }}" "{{ activemq_install_root }}/activemq"
  args:
    creates: "{{ activemq_install_root }}/activemq"
```

roles/internal/activemq/tasks/install.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- name: Create symlink
+  file:
+    state: link
+    src: "/opt/activemq/bin/activemq"
+    dest: "/etc/init.d/activemq"
+
+- name: Add activemq service to start up
+  shell: 'update-rc.d activemq defaults'
```
jonathangreen: This also needs to be made idempotent somehow. I would suggest dropping this entirely and handling it all using the ansible services module in the task below.

roles/internal/activemq/tasks/install.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Clean activemq path
+  file:
+    state: absent
+    path: "{{ activemq_install_root }}/activemq"
+
+- name: Download and unarchive activemq
+  unarchive:
```
jonathangreen: This also needs to be made idempotent somehow. Something like this would work: 

```yaml
- name: Download and unarchive activemq
  unarchive:
    remote_src: yes
    src: http://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz
    dest: "{{ activemq_install_root }}"
    owner: "{{ activemq_user }}"
    group: "{{ activemq_group }}"
    creates: "{{ activemq_install_root }}/activemq"
```

roles/internal/activemq/tasks/install.yml
```diff
@@ -0,0 +1,31 @@
...
+  file:
+    state: link
+    src: "/opt/activemq/bin/activemq"
+    dest: "/etc/init.d/activemq"
+
+- name: Add activemq service to start up
+  shell: 'update-rc.d activemq defaults'
+
+- name: Start activeMQ service
+  service:
```
jonathangreen: Using this module its easier to make the shell command above idempotent. 
http://docs.ansible.com/ansible/latest/service_module.html

The `enabled` flag should make this start at startup already.

Natkeeran:
@jonathangreen I've made most of the changes you had suggested.  It seems to be starting up, but I am not able to access the admin web console.  

I am not sure if this task is properly handled: https://github.com/Islandora-Devops/claw-playbook/blob/1901017f853ba0bfffcdc84a9e5b913b1ddadf8d/roles/internal/activemq/tasks/install.yml#L23

Received the following warning:
```
TASK [activemq : Start activeMQ service] ***************************************
 [WARNING]: The service (activemq) is actually an init script but the system is
managed by systemd
```
Status msg:
```
ubuntu@claw:/etc/init.d$ sudo service activemq status
● activemq.service - LSB: Starts ActiveMQ
   Loaded: loaded (/etc/init.d/activemq; bad; vendor preset: enabled)
   Active: active (running) since Thu 2017-08-31 15:25:20 UTC; 9min ago
     Docs: man:systemd-sysv-generator(8)
  Process: 29232 ExecStop=/etc/init.d/activemq stop (code=exited, status=0/SUCCESS)
  Process: 29365 ExecStart=/etc/init.d/activemq start (code=exited, status=0/SUCCESS)
    Tasks: 44
   Memory: 186.5M
      CPU: 4.400s
   CGroup: /system.slice/activemq.service
           └─29408 /usr/bin/java -Xms64M -Xmx1G -Djava.util.logging.config.file=logging.properties

Aug 31 15:25:20 claw systemd[1]: Starting LSB: Starts ActiveMQ...
Aug 31 15:25:20 claw activemq[29365]: INFO: Loading '/opt/activemq//bin/env'
Aug 31 15:25:20 claw activemq[29365]: INFO: Using java '/usr/bin/java'
Aug 31 15:25:20 claw activemq[29365]: INFO: Starting - inspect logfiles specified in logging.prope
Aug 31 15:25:20 claw activemq[29365]: INFO: pidfile created : '/opt/activemq//data/activemq.pid' (
Aug 31 15:25:20 claw systemd[1]: Started LSB: Starts ActiveMQ.
lines 1-18/18 (END)
```

jonathangreen:
@Natkeeran for me switching to the systemd module instead of the service module, gets rid of the warning you were seeing above. 

Something like this: 
```yaml
- name: Reload systemd
  systemd:
    daemon-reload: yes
    enabled: yes
    state: started
    name: activemq
```

When I reboot I see the service has started, and when I check its status I can see its started. The issue I believe with the admin page is that the ports for the admin page aren't forwarded in vagrant: 
https://github.com/Islandora-Devops/claw-playbook/blob/master/Vagrantfile#L26-L32

PR #7, opened 2017-08-29
Use Drupal JWT keys in Syn too

whikloj:
Simply Resolves https://github.com/Islandora-CLAW/CLAW/issues/702

This is very basic and only good for single server setups.

It will need more to do actual copying of a JWT if Fedora and Drupal are on separate machines.

### DIFF
--- a/inventory/vagrant/group_vars/webserver/general.yml
+++ b/inventory/vagrant/group_vars/webserver/general.yml
@@ -3,3 +3,5 @@
 webserver_app: yes
 openseadragon_iiiv_set_var: yes
 openseadragon_iiiv_server: http://localhost:8080/cantaloupe/iiif/2
+webserver_app_jwt_key_path: /home/ubuntu/auth
+
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -27,7 +27,7 @@ fcrepo_syn_sites:
     encoding: PEM
     anonymous: true
     default: true
-    path: /home/ubuntu/auth/public.key
+    path: "{{ webserver_app_jwt_key_path }}/public.key"
 
 fcrepo_syn_tokens:
   - user: admin


whikloj:
Probably easiest to test after https://github.com/Islandora-CLAW/CLAW/issues/690 and https://github.com/Islandora-CLAW/CLAW/issues/692 are merged.

DiegoPino:
@whikloj ok, will hurry up then.

inventory/vagrant/group_vars/webserver/general.yml
```diff
@@ -1,3 +1,5 @@
 ---
 
 webserver_app: yes
+webserver_app_jwt_key_path: /home/ubuntu/auth
```
DiegoPino: I saw a few comments about allowing multi platform approaches for our ansible playbook. Where would this one be if using Centos/Redhat?

whikloj: I guess it would be where ever you wanted to store these keys. Could be `/etc/ssl` if you so choose, but you'd have to make that directory. This directory is generated when the SSL keys are made, so all I am doing is copying the variable up to a higher point to allow both roles to access it the same.

DiegoPino: Then i guess we need to make this platform independent, you run Redhat right? 
This here is just for ubuntu:
https://github.com/Islandora-Devops/claw-playbook/blob/3efb1326addd71f0933d7945ea1bddf15cd0e7cc/inventory/vagrant/group_vars/tomcat.yml#L29
Mostly because every sprint issue we have right now says we should
try supports both Ubuntu and CENTOS. Just a comment because seems like that is one of the advantages of ansible at all (other than learning a new pseudo language we will forget if not using it on a daily basis)

DiegoPino: @jonathangreen @whikloj probably 
```YML
webserver_app_jwt_key_path:  {{ fcrepo_syn_sites[path] | default(/home/ubuntu/auth) }}
```
Not sure about the fcrepo_syn_sites: path: nomenclature (way of accesing) but i would for sure think we should reuse as much vars as possible

dannylamb:
@whikloj Whenever you get a chance, can you try rebasing this?  I'll gladly test. Once this is done, we're at feature parity (and a bit more, I'd say) and can deprecate claw_vagrant :tada:

whikloj:
Rebasing and testing

whikloj:
This PR was replaced by https://github.com/Islandora-Devops/claw-playbook/pull/25

PR #8, opened 2017-08-29
API-X and Alpaca Ansible roles WIP

DiegoPino:
Please don't review and don't comment yet (well you can if you want). 
This is just a way of forcing myself to commit code while I test. 
Ansible Syntax checks are passing, but I need to deal with ```Gradle``` output, idempotency and checking if the compiling actually worked out.
 Also, happens that I feel I'm using like 10% of Ansible's capabilities 😢 

Will update this pull request message once it is usable.

Should resolve https://github.com/Islandora-CLAW/CLAW/issues/692

### DIFF
--- a/karaf.yml
+++ b/karaf.yml
@@ -5,3 +5,4 @@
 
   roles:
     - karaf
+    - apix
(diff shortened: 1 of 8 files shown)


jonathangreen:
I'm going to close this PR, there is a new one incoming that builds on this one.

PR #10, opened 2017-08-30
Enable members block for collections

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/706


jonathangreen:
`vagrant up`

PR #11, opened 2017-08-30
Updated README.md and created new index.md

g7morris:
Hello Team, 

I've updated the README.md and have created a stub index.md for more details which I hope to fill prior to Friday. 

@jonathangreen @dannylamb Style okay? Too much, too little. 

Thanks, 
Gavin

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,37 +3,102 @@
 
 ## Introduction
 
-The is an Ansible playbook for Islandora CLAW. It also has a vagrant file to bring up a development
-environment virtual machine for Islandora CLAW.
+This is the development environment virtual machine (vm) for running Islandora 7.x-2.x (a.k.a. [Islandora CLAW](https://github.com/Islandora-CLAW/CLAW)). 
 
-This virtual machine **should not** be used in production **yet**.
+_This virtual machine **should not** be used in production **yet**._
+
+* It should work on any operating system that supports Ansible, Git, VirtualBox and Vagrant. 
+
+* It uses a suite of [Ansible](https://www.ansible.com/how-ansible-works) playbooks called within the Vagrantfile's provisioners section to install [Islandora CLAW](https://github.com/Islandora-CLAW/CLAW), its dependencies, additional software, user roles and more.
+
+* The information listed below is to serve as a quick getting started guide. 
+  * For more detailed information please navigate to the Technical documentation [here](/docs/index.md)
+
+---
 
 ## Requirements
 
-1. [VirtualBox](https://www.virtualbox.org/)
-2. [Vagrant](http://www.vagrantup.com/) 1.8.5+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -3,37 +3,102 @@
...
+---
 
 ## Requirements
 
-1. [VirtualBox](https://www.virtualbox.org/)
-2. [Vagrant](http://www.vagrantup.com/) 1.8.5+
-3. [git](https://git-scm.com/)
-4. [ansible](https://www.ansible.com/community) 2.3+
+1. [VirtualBox](https://www.virtualbox.org/wiki/Downloads) 5.1+
+2. [Vagrant](https://www.vagrantup.com/downloads.html) 1.95+
```
MarcusBarnes: Assuming that 1.95+ here is supposed to be 1.9.5+.

g7morris: @MarcusBarnes Good catch! that should be 1.9.5+ for sure. Thanks!

README.md
```diff
@@ -3,37 +3,102 @@
...
+## System Ports
+The following services and ports are mapped/exposed from the development environment virtual machine (guest) to the (host) laptop, workstation or server that runs Virtualbox.
+
+Please note that in the event of conflict, the port values for the **host ONLY** should be changed within the Vagrantfile. These values can be arbitrary. This can often occur when running more than one Vagrant server at time. 
+
+**Original Vagrant port settings**
+* `guest: 8000, host: 8000` # **Apache**
+* `guest: 8080, host: 1234` # **Tomcat**
+* `guest: 8181, host: 8181` # **Karaf**
+* `guest: 8282, host: 8282` # **Islandora Microservices**
```
dannylamb: This line can be removed

dannylamb: Or more accurately, should be replaced with

`guest: 8081, host: 8081` # **Api-X**

docs/index.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Installation 
+
+The [installation section](install/README.md) provides and overview on how to create a virtual development environment.
+
+## Using Ansible 
+
+* List of commands here
+
+## Ansible Playbook structure
```
jonathangreen: I think this section goes a little too detailed. I think the directory structure of the repo changes fairly quickly, and including it here assures that it will be outdated soon.

jonathangreen:
@g7morris other then the couple comments here, I think the style of this is good and its always great to have some more docs.

DonRichards:
This should probably get pushed through before the sprint. Just a suggestion. This does look like a good start and if @g7morris doesn't have the time to make the suggested modifications I think it would be useful to push it through as it sits. This would be a good starting point.

dannylamb:
Out of date, sadly.  Closing

PR #12, opened 2017-09-13
Add activemq config for fcrepo

jonathangreen:
## Ticket
https://github.com/Islandora-CLAW/CLAW/issues/711

## Issue
Fcrepo wasn't coming up because of two activemq, embedded and external. Same fix as for the vagrant box we need to configure fedora to not start activemq broker.

## Testcase
Run playbook and make sure fcrepo comes up.

### DIFF
--- a/roles/internal/fcrepo/defaults/main.yml
+++ b/roles/internal/fcrepo/defaults/main.yml
@@ -4,4 +4,4 @@ fcrepo_data_dir: /var/lib/tomcat8/fcrepo4-data
 fcrepo_war_path: "{{ tomcat8_home }}/webapps/fcrepo.war"
 fcrepo_home_dir: /opt/fcrepo
 fcrepo_activemq_template: activemq.xml.j2
-fcrepo_tomcat_config_dir: "{{ tomcat8_home }}/webapps/fcrepo/WEB-INF/classes/config/"
+fcrepo_config_dir: "{{ fcrepo_home_dir }}/configs"
--- a/roles/internal/fcrepo/defaults/main.yml
+++ b/roles/internal/fcrepo/defaults/main.yml
@@ -3,3 +3,5 @@ fcrepo_user: tomcat8
 fcrepo_data_dir: /var/lib/tomcat8/fcrepo4-data
 fcrepo_war_path: "{{ tomcat8_home }}/webapps/fcrepo.war"
 fcrepo_home_dir: /opt/fcrepo
+fcrepo_activemq_template: activemq.xml.j2
+fcrepo_tomcat_config_dir: "{{ tomcat8_home }}/webapps/fcrepo/WEB-INF/classes/config/"
--- a/roles/internal/fcrepo/templates/activemq.xml.j2
+++ b/roles/internal/fcrepo/templates/activemq.xml.j2
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+    xmlns:amq="http://activemq.apache.org/schema/core"
+    xmlns:context="http://www.springframework.org/schema/context"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+    http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.0.xsd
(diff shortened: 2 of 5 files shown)


jonathangreen:
@whikloj @dannylamb @ajs6f 

ran into this while testing the crayfish pull. if anyone has time to review and merge would be most appreciated.

roles/internal/fcrepo/defaults/main.yml
```diff
@@ -3,3 +3,5 @@ fcrepo_user: tomcat8
 fcrepo_data_dir: /var/lib/tomcat8/fcrepo4-data
 fcrepo_war_path: "{{ tomcat8_home }}/webapps/fcrepo.war"
 fcrepo_home_dir: /opt/fcrepo
+fcrepo_activemq_template: activemq.xml.j2
+fcrepo_tomcat_config_dir: "{{ tomcat8_home }}/webapps/fcrepo/WEB-INF/classes/config/"
```
ajs6f: It's generally not best practice to put config like this inside the filesystem space of the webapp itself... maybe someplace inside `/usr/local` or `/etc` or the like?

roles/internal/fcrepo/templates/activemq.xml.j2
```diff
@@ -0,0 +1,28 @@
...
+    http://activemq.apache.org/schema/core http://activemq.apache.org/schema/core/activemq-core.xsd">
+
+    <context:property-placeholder/>
+
+    <!--
+        The <broker> element is used to configure the ActiveMQ broker.
+    -->
+    <broker xmlns="http://activemq.apache.org/schema/core" brokerName="localhost">
+      <persistenceAdapter>
+        <kahaDB directory="${fcrepo.activemq.directory:ActiveMQ/kahadb}"/>
```
ajs6f: Are we setting `fcrepo.activemq.directory` somewhere?

roles/internal/fcrepo/defaults/main.yml
jonathangreen: I agree with you 100%. Fedora itself puts a `activemq.xml` file at this path, that I am trying to override here. Is there some way to pass a path for this config to fedora so it reads the alternative configuration? I'm not sure how I would do that.

roles/internal/fcrepo/templates/activemq.xml.j2
jonathangreen: I assume its being set in the fedora configs somewhere. I stole this line directly from the activemq config that ships with fcrepo. 


https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-configs/src/main/resources/config/activemq.xml#L78-L80

ajs6f: Okay, cool, then it's not CLAW's problem.

roles/internal/fcrepo/defaults/main.yml
ajs6f: Jeez, I can't believe Fedora does that. No, I can totally believe that. If that's the way that Fedora does it, I guess we're probably better off leaving it alone.

ajs6f:
I'll wait a bit through today in case anyone else has comments, then I'll merge this. It's pretty much a no-brainer.

roles/internal/fcrepo/defaults/main.yml
whikloj: I think we reverted my work on Fedora to make a single configuration file that you could override with a JAVA_OPT. I'll look back and see.

whikloj: Ok, I think we can put the `activemq.xml` anywhere and set use this https://github.com/fcrepo4/fcrepo4/blob/fcrepo-4.7.4/fcrepo-webapp/src/main/resources/spring/jms.xml#L18-L19

whikloj: So I'm not sure how we do this in Ansible, but we need to add
`-Dfcrepo.activemq.configuration=file:/blah/blah/activemq.xml`
to
https://github.com/Islandora-Devops/claw-playbook/blob/master/inventory/vagrant/group_vars/tomcat.yml#L9-L22

jonathangreen: Cool. Thanks @whikloj I'll update the playbook.

jonathangreen:
@whikloj @ajs6f updated the path for the config file.

ajs6f:
I know @dannylamb is out this week, so can I assume that everyone who wants to has chimed in on this PR?

PR #13, opened 2017-09-15
Alpaca, API-X and Karaf

jonathangreen:
## Issue
This is an extension of the work done by @DiegoPino here:
https://github.com/Islandora-Devops/claw-playbook/pull/8

Resolves:
https://github.com/Islandora-CLAW/CLAW/issues/692

## Information
This PR updates the Karaf role to add two new Ansible modules for manipulating Karaf repos and Karaf features. I am usually fairly opposed to adding new modules, since they increase complexity, you both have to understand the playbook and the python code, but I was having a lot of trouble creating tasks to add karaf repos and features in a nice idempotent way, so I went the module route. I think they will come in handy as we do more work with Karaf.

This PR then adds two new roles, API-X and Alpaca that leverage these new modules to setup API-X and Alpaca in Karaf. 

## Testing
Bring up vagrant with claw playbook and make sure that: 
- Karaf is installed and working
- API-X features are deployed and working
- Alpaca features are deployed and working

## Interested parties
@DiegoPino @whikloj @ajs6f @dannylamb


roles/internal/alpaca/defaults/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+alpaca_features:
+  - islandora-http-client
+  - islandora-connector-broadcast
+  - islandora-indexing-triplestore
+  - islandora-indexing-fcrepo
+
+alpaca_karaf_dir: /opt/karaf
+alpaca_karaf_etc_dir: "{{ alpaca_karaf_dir }}/etc"
+alpaca_local_mvn_path: /home/{{ alpaca_user }}/.m2/repository
```
ajs6f: It might be a little more Maven-idiomatic to do this with a templated [`settings.xml`](https://maven.apache.org/settings.html) file.

dannylamb:
I tried running this and I got the following:
```
TASK [alpaca : Add Alpaca Features] ********************************************
changed: [default] => (item=islandora-http-client)
changed: [default] => (item=islandora-connector-broadcast)
changed: [default] => (item=islandora-indexing-triplestore)
failed: [default] (item=islandora-indexing-fcrepo) => {"failed": true, "item": "islandora-indexing-fcrepo", "msg": "client: JAVA_HOME not set; results may vary\nError executing command: Error restarting bundles\n"}
	to retry, use: --limit @/home/daniel/Code/Environments/claw-playbook/playbook.retry

PLAY RECAP *********************************************************************
default                    : ok=295  changed=156  unreachable=0    failed=1   

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

Pretty sure JAVA_HOME is getting set somewhere, but maybe we're not doing it for the Karaf user?

jonathangreen:
@dannylamb `JAVA_HOME ` isn't set, but Karaf complains about this all the time, but still works. I am seeing this error intermittently when I've been testing, it looks like its a failure of the `islandora-indexing-fcrepo` bundle to deploy. 

Perhaps it is related to this issue https://github.com/Islandora-CLAW/CLAW/issues/686, only here we check the return code of all commands, so we see the failures occurring more clearly. 

Seems like its working for just fine every few builds...

DiegoPino:
@jonathangreen same issue (ENV) I had during my humble attempts First i tried setting it via http://docs.ansible.com/ansible/latest/playbooks_environment.html and fetching it back with  karaf_home: "{{ lookup('env','KARAF_HOME')| default('/opt/karaf') }}" and no luck (did not seem to stick)
Wonder if we could get:
Could we get KARAF_HOME and JAVA_HOME globally set for both, the Ubuntu, and whichever user Ansible uses?

jonathangreen:
@DiegoPino I believe `JAVA_HOME` not being set isn't especially important in this case, as it seems like Karaf does the right thing anyway.

roles/internal/alpaca/defaults/main.yml
```diff
@@ -0,0 +1,29 @@
...
+alpaca_user: karaf
+
+alpaca_settings:
+  - pid: ca.islandora.alpaca.connector.broadcast
+    settings:
+      input.stream: broker:queue:islandora-connector-broadcast
```
dannylamb: Should be `activemq:queue:islandora-connector-broadcast`.  Anytime you see `broker:*` it's just a placeholder for `activemq` or whatever other queuing technology camel supports.

roles/internal/alpaca/defaults/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  - pid: ca.islandora.alpaca.indexing.triplestore
+    settings:
+      error.maxRedeliveries: 10
+      input.stream: activemq:queue:islandora-indexing-triplestore
+      triplestore.baseUrl: http://localhost:8080/bigdata/namespace/kb/sparql
```
dannylamb: We'll also need

```yml
- pid: org.fcrepo.camel.indexing.triplestore.cfg
 settings:
   input.stream: activemq:queue:islandora-connector-broadcast
   triplestore.reindex.stream: activemq:queue:triplestore.reindex
   triplestore.baseUrl: http://localhost:8080/bigdata/namespace/kb/sparql
```

roles/internal/apix/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
...
+  - fcrepo-api-x
+  - fcrepo-indexing-triplestore
+  - fcrepo-service-camel
+
+apix_karaf_etc_dir: "{{ karaf_install_symlink }}/etc"
+
+apix_config:
+  - pid: org.fcrepo.apix.registry.http
+    settings:
+      timeout.socket.ms: 1000
```
dannylamb: Sorry, put this at the wrong spot earlier.

We'll also need

```yml
- pid: org.fcrepo.camel.indexing.triplestore.cfg
 settings:
   input.stream: activemq:queue:islandora-connector-broadcast
   triplestore.reindex.stream: activemq:queue:triplestore.reindex
   triplestore.baseUrl: http://localhost:8080/bigdata/namespace/kb/sparql
```

dannylamb:
@jonathangreen Ha yeah, weird.  So looks like if one of the features fails to deploy, we get the JAVA_HOME warning even though that's not the actual reason for the failure.  ¯\_(ツ)_/¯

I pushed through with a retry and found a few config values that need to be set.  The Api-X proxy endpoint wasn't available either, but that may be because of those missing values.  I'm not sure.

dannylamb:
@jonathangreen I'm giving this another whirl now that Islandora-CLAW/CLAW#686 has landed.

roles/internal/apix/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  - pid: org.fcrepo.camel.indexing.triplestore.cfg
```
dannylamb: So for some reason, this is creating a second config file with an extra `.cfg` at the end: `org.fcrepo.camel.indexing.triplestore.cfg.cfg`.  If we can sort this out, we're good to go.  All other functionality is there.  The proxy is up and Fedora + 3store sync'ing is happening from Drupal, just this last bit is preventing the 3store indexer from Fedora..

roles/internal/apix/tasks/config.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- name: Template settings
+  template:
+    src: apix.cfg.j2
+    dest: "{{ apix_karaf_etc_dir }}/{{ item.pid }}.cfg"
```
dannylamb: Oh, I see.  It's here.  Either drop the .cfg's in the pids or drop it here and put them all the pid entries at `roles/internal/apix/defaults/main.yml`

PR #15, opened 2017-09-20
Use snapshots instead of compiling by default

jonathangreen:
## Github issue

Related to: 
https://github.com/Islandora-CLAW/Alpaca/pull/45

Issue: 
https://github.com/Islandora-CLAW/CLAW/issues/713

## Changes

Instead of compiling Alpaca by default, just bring up the box with the snapshot.

## Testing

Bring up box and make sure everything still works, and that Alpaca comes out of snapshots in maven.


dannylamb:
`vagrant up`

PR #17, opened 2017-09-21
Changing fcrepo-indexing-triplestore to read from Fedora's default topic

dannylamb:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/719

Testing instructions:
In `/opt/karaf/etc/org.fcrepo.camel.indexing.triplestore.cfg`, `input.stream` should be `activemq:topic:fedora`, not `activemq:queue:islandora-connector-broadcast`.  If you're paranoid, make an object in Drupal.  Both Fedora sync'ing AND 3-store indexing from Fedora should both be functioning.  Because of the mix-up this PR resolves, there were timing issues on who consumed from the queue, so you'd get either one or the other, but never both.

### DIFF
--- a/roles/internal/apix/defaults/main.yml
+++ b/roles/internal/apix/defaults/main.yml
@@ -14,6 +14,6 @@ apix_config:
       timeout.socket.ms: 1000
   - pid: org.fcrepo.camel.indexing.triplestore
     settings:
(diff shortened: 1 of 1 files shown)


dannylamb:
^^ @Islandora-CLAW/committers

dannylamb:
@whikloj @jonathangreen @DiegoPino

PR #20, opened 2017-10-13
Fix incorrect path for MySQL SQL file.

jonathangreen:
## Reason for PR

In https://github.com/Islandora-Devops/claw-playbook/pull/14 it looks like I broke the path for the MySQL sql file, so postgresql is working in the playbook but not mysql. Issue found by @whikloj.

## Testing

Run playbook and make sure gemini sets up its MySQL database correctly.


whikloj:
👍

PR #25, opened 2017-11-08
Copy JWT around

whikloj:
Partially resolves: Islandora-CLAW/CLAW#702

This PR adds a new role called **keymaster**. All it does is generate a key in a single location and allow for the copying of either or both of the private/public keys around.

This role is not referenced in the playbook except by other roles that require the keys.

To test, please pull this PR and build it.

Everything should work out of the box.

### Thoughts of things not yet done: 
1. I haven't added a way to copy the keys in from the host to the system.
1. If this is on a multiple server setup, perhaps the key generation should be on the host system and then could be copied to all the various remote targets.

Vinz Clortho

### DIFF
--- a/roles/internal/crayfish/defaults/main.yml
+++ b/roles/internal/crayfish/defaults/main.yml
@@ -6,6 +6,8 @@ crayfish_services:
   - Houdini
   - Milliner
   - Hypercube
+  
+crayfish_install_dir: /var/www/html/Crayfish
 
 crayfish_syn_token: islandora
 
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -22,12 +22,14 @@ tomcat8_java_opts:
   - -Dcantaloupe.config={{ cantaloupe_symlink }}/cantaloupe.properties
   - -Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true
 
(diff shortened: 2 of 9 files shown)


DiegoPino:
@whikloj: first, this is just a question so dismiss if you don't have time.
Does JWT key generation work like an ssh key? if so, Does this module/or similar could help with  http://docs.ansible.com/ansible/latest/authorized_key_module.html unresolved 1.? and 2.? 
Probably i don't even know what i'm talking about... but it made some sense a few minutes ago

DiegoPino:
or something like (for 2.)
https://gist.github.com/fulv/3928d098e8c35af1cc5363a4d2d4fcd0

whikloj:
@DiegoPino the number 1 is more when you have multiple separate target servers configured. Currently this works to copy the SSL keys from a single location to the various services. This makes it so they all have a copy of the required private or public key. 

So this isn't really about SSH keys used for access a server or at least I don't think it it. I'll try to look more closely at what you linked to.

dannylamb:
@whikloj Spinning this up now to test.

roles/internal/crayfish/defaults/main.yml
```diff
@@ -6,6 +6,8 @@ crayfish_services:
   - Houdini
   - Milliner
   - Hypercube
+  
+crayfish_install_dir: /var/www/html/Crayfish
```
dannylamb: Thanks for this

dannylamb:
This indeed copies the keys out and wires up the microservices.  Made some objects and they got indexed in the triplestore and fedora.

PR #27, opened 2017-11-09
Handing Centos setup in bootstrap.yml

dannylamb:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/749

Depends on https://github.com/Islandora-CLAW/CLAW/issues/746

This PR sets up bootstrapping for Centos systems.  It skips installing python and aptitude, and uses a different list of packages.  I'm replacing `build-essential` with equivalent packages available to yum, and am removing some that don't exist without adding extra repos.

You can test it with
```bash
ISLANDORA_VAGRANT_DISTRO="centos/7" vagrant up --no-provision
ansible-playbook -i inventory/vagrant -t bootstrap playbook.yml
```

It works, but I'm sure it can be done better and am open to suggestions.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -3,10 +3,8 @@
 - name: bootstrap
   hosts: all
   become: yes
-  gather_facts: false
+  gather_facts: true
 
-  # TODO: this probably needs to be made general for RHEL, it will only
-  # work for ubuntu right now.
   tasks:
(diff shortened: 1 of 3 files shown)


DigitLib:
WORK!!!

whikloj:
I don't have a free Redhat/Centos box sitting around. But this looks fine and if @DigitLib says its good then I'm good. 👍

PR #29, opened 2017-11-10
Issue 752

dannylamb:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/752
Supersedes: https://github.com/Islandora-Devops/claw-playbook/pull/26

Fixes Ubuntu after breaking it with https://github.com/Islandora-Devops/claw-playbook/pull/27.

This passes the environment variable for what linux distribution you want as an "extra_var" so it's available in bootstrap.yml.  This way we don't have to gather facts, which was breaking because ubuntu doesn't ship with python.

### To test
`vagrant up` and if you make it past bootstrapping, you're :+1:


MarcusBarnes:
I'll test now.  Thanks.

dannylamb:
@MarcusBarnes++

MarcusBarnes:
The Ubantu VM was successfully built.

bootstrap.yml
```diff
@@ -14,7 +14,7 @@
       changed_when:
         - output.stdout != ""
         - output.stdout != "\r\n"
-      when: ansible_os_family == "Debian"
+      when: islandora_vagrant_distro == "ubuntu/xenial64"
```
jonathangreen: To keep the play from failing when this isn't defined. Say not run with vagrant, I would do something like: 
```yaml 
when: islandora_vagrant_distro|default('false') == "ubuntu/xenial64"
```

So there is a default value. And don't install python with apt by default.

bootstrap.yml
```diff
@@ -14,7 +14,7 @@
...
```
jonathangreen: Then I think I would add a new task here to gather facts: 
```yaml
# Gathers facts, since it has to be disabled for ubuntu above
- name: Gather Facts
  setup:
```

bootstrap.yml
```diff
@@ -23,20 +23,20 @@
         name: aptitude
         state: present
         update_cache: yes
-      when: ansible_os_family == "Debian"
+      when: islandora_vagrant_distro == "ubuntu/xenial64"
```
jonathangreen: Then I would change these back to the more sane ansible defaults.

bootstrap.yml
Natkeeran: I would suggest that we provide a default value for `islandora_vagrant_distro` variable such as `ubuntu/xenial64`.

Also a more generic variable name is more appropriate, `islandora_distro`.

dannylamb:
Sure, these all sound like good improvements. It'll be a bit before I can get to them. Stay tuned.

dannylamb:
@Natkeeran @MarcusBarnes @jonathangreen I've just updated the PR.   Take a look when you have a chance.

PR #33, opened 2017-11-19
Replace deprecated include with import_playbook

fkleon:
**GitHub Issue**: Islandora-CLAW/CLAW#734

A start on fixing the deprecation warnings that are displayed when running Ansible 2.4.

This PR fixes the following deprecation warning:
```
[DEPRECATION WARNING]: 'include' for playbook includes. You should use 
'import_playbook' instead. This feature will be removed in version 2.8. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ If you struggle with following the instructions for setting up the requirements 
 1. [VirtualBox](https://www.virtualbox.org/)
 2. [Vagrant](http://www.vagrantup.com/) (Known to work with 2.0.1 - 2.1.2)
 3. [git](https://git-scm.com/)
-4. [ansible](https://www.ansible.com/community) 2.3+
+4. [ansible](https://www.ansible.com/community) 2.4+
 5. [virtualbox-vbguest](https://github.com/dotless-de/vagrant-vbguest) plugin (If targeting CENTOS)
 
 ## Variables
--- a/playbook.yml
+++ b/playbook.yml
@@ -1,12 +1,12 @@
 ---
 
-- include: bootstrap.yml
+- import_playbook: bootstrap.yml
   tags:
     - bootstrap
-- include: database.yml
-- include: solr.yml
-- include: webserver.yml
-- include: tomcat.yml
-- include: crayfish.yml
-- include: karaf.yml
-- include: post-install.yml
+- import_playbook: database.yml
+- import_playbook: solr.yml
+- import_playbook: webserver.yml
+- import_playbook: tomcat.yml
+- import_playbook: crayfish.yml
+- import_playbook: karaf.yml
+- import_playbook: post-install.yml


dannylamb:
@fkleon This would break for those who are running Ansible 2.3.  If we can confirm we're all good on 2.4, then we'll need to update the readme to advertise that you'll need to use 2.4 only.

It would also be good to try and enforce this restriction within Ansible itself, but that can be handled in a separate PR.  Looks like drupal-vm is handling it [this way](https://github.com/kekkis/drupal-vm/commit/f1708b86692675bd296df7d4ea6e30a369c9a5c4).

fkleon:
@dannylamb You're absolutely right, and there should be consensus about the supported Ansible versions before merging this. I've seen the discussion on the issue tracker about enforcing a particular Ansible version, and using a `pre_task` for that purpose seems perfectly fine to me.

Another option to be fully independent of the host Ansible version might be to use the [ansible_local](https://www.vagrantup.com/docs/provisioning/ansible_local.html) provisioner, and bootstrap Ansible on the guest through the `install` option of that provisioner, or a simple shell script.

whikloj:
Can we (and do we want to) compare against the current version of Ansible and have a pre-2.4 and post-2.4 version of the playbook.yml to run?

rosiel:
I came here cause i'm still getting these messages... the warnings now say 

> [DEPRECATION WARNING]: 'include' for playbook includes. You should use 
'import_playbook' instead. This feature will be removed in version 2.8. 
Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.
 This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is
 discouraged. The module documentation details page may explain more about this
 rationale.. This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: The use of 'static' has been deprecated. Use 
'import_tasks' for static inclusion, or 'include_tasks' for dynamic inclusion. 
This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.

And because Ansible 2.7 has been out since October, and with a 3x/year release schedule (based on past releases) we can probably expect 2.8 in a few months, it might be worth looking at this sooner than later to prevent noob frustration.

whikloj:
@fkleon sorry for the looooooong delay, if you could rebase this to resolve the conflicts I will make time to review and get it in.

DonRichards:
As long as there isn't a platform where a stable version at or above 2.4 isn't currently already available . I think it's safe to move forward with this.

fkleon:
@whikloj No worries. I have rebased this PR and given it a very quick test with Ansible 2.5.1 (as packaged by Ubuntu 18.04).

seth-shaw-unlv:
The import_playbook changes were included on dev with d0f7670 (while this PR targets main). The only remaining change is the ansible version bump in the README. Do we still need this PR?

PR #34, opened 2017-11-23
Define fedora.home environment variable

whikloj:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/757

For some reason with all the https://github.com/Islandora-CLAW/CLAW/issues/640 PRs merged Fedora started to have permissions trouble. The simplest answer seems to be to define where fedora should store its data (and the problematic velocity.log file).

While this is one not a required parameter for Fedora to run, it otherwise uses the current working directory which I think is `/var/lib/tomcat8`. So I am guessing there are some issues using that.

For more fedora information see the [Application Configuration](https://wiki.duraspace.org/display/FEDORA4x/Application+Configuration) documentation.

### DIFF
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -16,6 +16,7 @@ tomcat8_java_opts:
   - -XX:MaxNewSize=256m
   - -XX:PermSize=256m
   - -XX:MaxPermSize=256m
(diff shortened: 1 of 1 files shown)


whikloj:
@dannylamb yeah this can go in now. claw-playbook will still be broken until we get the Alpaca release done, but Fedora will come up.

PR #36, opened 2017-11-24
Ensure islandora namespace is in blazegraph

Natkeeran:
In Ansible, handlers get fired last.  If there are tasks that need the handlers to be fired, then handlers must be flushed (`- meta: flush_handlers`).  Initially, the Islandora namepsace was not being applied to blazegraph, perhaps because the tasks get run before blazegraph comes up.  This PR ensures that blazegraph comes up before running those tasks.  

## How should this be tested?
* Get the PR
* Ensure that the islandora namespace is available in http://localhost:8080/bigdata/#namespaces

## Interested parties
@dannylamb 
@jonathangreen

### DIFF
--- a/roles/internal/alpaca/defaults/main.yml
+++ b/roles/internal/alpaca/defaults/main.yml
@@ -27,4 +27,4 @@ alpaca_settings:
     settings:
       error.maxRedeliveries: 10
       input.stream: activemq:queue:islandora-indexing-triplestore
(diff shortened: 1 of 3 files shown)


MarcusBarnes:
I've tested.  The VM came up and the islandora namespace was in the list of available namespaces in Blazegraph.

dannylamb:
`vagrant up`

Natkeeran:
@dannylamb right, that why I did not see anything get populated into islandora namespace.  Thanks.  Made the changes.

PR #39, opened 2017-11-26
ufw install and apply rules

Natkeeran:
* This PR installs ufw firewall and adds basic rules.  

## How should this be tested?
* Get PR
* SSH into vagrant, check ufw status: `sudo ufw status`.  Should say it is active and list the allowed ports.
* CRUD tests
* This need to be tested in all supported environments 

## Interested parties
@dannylamb @whikloj @jonathangreen

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -41,3 +41,7 @@
 - src: geerlingguy.drupal-console
   name: drupal-console
   version: 1.1.0
+
+- src: franklinkim.ufw
+  name: ufw
+  version: 1.6.0
--- a/roles/internal/ufw-rules/defaults/main.yml
+++ b/roles/internal/ufw-rules/defaults/main.yml
@@ -1,0 +1,24 @@
+ufw_default_incoming_policy: deny
+ufw_default_outgoing_policy: allow
+
+ufw_rules:
+  - rule: allow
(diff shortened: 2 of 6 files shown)


jonathangreen:
I can totally see how this would be handy on a real server, but I can also see it being a bit of a pain in the butt when doing development on a throwaway vagrant box. Maybe we could wrap the role install in a `when:` statement, so it can be disabled?

Natkeeran:
@jonathangreen Yes, agree with that approach.  Should I check for a specific variable in the when statement such as install_ufw?

jonathangreen:
Yes I like that. Maybe something like: 
```yaml
  roles:
      - role: ufw-rules
        when: install_ufw
```

Natkeeran:
I separated the role: https://github.com/Natkeeran/ansible-role-ufw

It has ` franklinkim.ufw` as a dependency.

@dannylamb  @jonathangreen @whikloj 

Should we make this an optional role?

whikloj:
@Natkeeran as this is now an external role, do you want to add some information to the docs on how you could add and use it to secure your and close this?

Natkeeran:
This is an external optional role.  Closing this PR.

PR #41, opened 2017-12-07
No Ubuntu Required

whikloj:
* Remove dependency on ubuntu user
* Fix Ubuntu bootstrap

https://www.youtube.com/watch?v=r0qBaBb1Y-U

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -9,7 +9,7 @@
     # python isn't installed by default on ubuntu 16.04 so we need
     # to do that with the raw command before
     - name: install python
-      raw: test -e /usr/bin/python || apt-get install -y python
+      raw: test -e /usr/bin/python || (apt-get update; apt-get install -y python;)
       register: output
       changed_when:
         - output.stdout != ""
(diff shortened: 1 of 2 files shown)


MarcusBarnes:
Appears to build successfully.  The Islandora namespace in Blazegraph was not selected by default, but this may not related to the changes in this pull-request.  

@whikloj Any other things I should check for when testing and reviewing this pull-request?  Thank you in advance.

whikloj:
Not really, this was just to ensure that the system worked without relying on the `/home/ubuntu` directory.

dannylamb:
@MarcusBarnes Turns out setting that default namespace for the web UI is a bit of an ordeal: https://github.com/Islandora-CLAW/CLAW/issues/536

We settled on adding documentation around it in https://github.com/Islandora-CLAW/claw_vagrant/pull/54

It's probably worth adding that to claw-playbook documentation as well.

MarcusBarnes:
@dannylamb Thanks for the clarification.

whikloj:
Oh 🤦‍♂️  I was thinking of @Natkeeran's work to add the islandora namespace...I wasn't thinking about it being auto-selected in Blazegraph. Yeah @MarcusBarnes that seems to be a browser/cookies thing and not easy to affect.

PR #42, opened 2017-12-08
Add a config to disable selinux

whikloj:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/759

If there is a cleaner way to do this, please let me know. I'm sure we could add some selinux config to allow this port but for now it seems easier to just skip it all together.

### DIFF
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -44,3 +44,20 @@
         state: present
       with_items: "{{ islandora_extra_centos_packages }}"
       when: ansible_os_family == "RedHat"
+
+    # Check Selinux
+    - name: check selinux status
+      raw: sestatus
+      register: sestatus_out
+      when: ansible_os_family == "RedHat"
+
+    # Parse sestatus down to one of two options.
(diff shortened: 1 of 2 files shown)


whikloj:
Sorry this is for Centos only, so you want to pull this PR and run
```
ISLANDORA_DISTRO="centos/7" vagrant up
```

You'll still fail but at PHP which is a separate issue 😉

seth-shaw-unlv:
As a data point, I've been using this PR is all of my CentOS testing and it has worked fine.

DigitLib:
I also can confirm that tihs PR working fine on my CentOS testing. For PHP issue there is geerligguy remo repo role which install desirable version (5.6, 7.0, 7.1, 7.2) I tested it succesfully but there is a question about additional php installation as defined in inventory/vagrant/group_vars/webserver/php.yml. On CentOS we need only php-mysql and php-pgsql and it installed it sucessfuly. When I solve a multi OS definition of additional phps I'll pull it for testing.

PR #47, opened 2018-01-12
set default drupal mysql user as drupal8

Natkeeran:
Several users have reported that they are running into issue related to this PR https://github.com/Islandora-Devops/claw-playbook/pull/45.  

Though I cannot fully reproduce with vagrant, I think it may be due to mysql trying to create root user.  I suppose the conditional statement does not get applied in the variable setting.

It is good practice to use non root user of the database.  Thus, changing the default drupal user from root to drupal8.  

## How should this be tested?
vagrant up

## Interested parties
@jonathangreen @dannylamb

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ The default Drupal login details are:
 
 ### MySQL
   
-  * username: root
+  * username: drupal8
   * password: islandora
 
 ### Fedora4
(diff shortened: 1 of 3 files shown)


seth-shaw-unlv:
Works for me when provisioning on CentOS.

PR #50, opened 2018-01-23
Can't conditionally define a variable

whikloj:
Resolves: https://github.com/Islandora-CLAW/CLAW/issues/778

This error only occurs if you try to run the playbook a second time.
So if you bring up a vagrant with claw-playbook and then run (this assumes Ubuntu)
```
ansible-playbook -i inventory/vagrant -u ubuntu -e islandora_distro="ubuntu/xenial" playbook.yml
```
you should see the error documented in the ticket above.

An alternative would be to include a variable file using the `when:` statement but I think it is probably better to get people used to checking the variable files when altering the configuration.

Pinging @Natkeeran as he has used non-root MySQL users.

### DIFF
--- a/inventory/vagrant/group_vars/all.yml
+++ b/inventory/vagrant/group_vars/all.yml
@@ -26,9 +26,13 @@ islandora_extra_centos_packages:
 
 postgresql_user: postgres
 
-mysql_users:
-  - name: "{{ drupal_db_user }}"
-    host: "%"
-    password: "{{ drupal_db_password }}"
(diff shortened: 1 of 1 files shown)


Natkeeran:
@whikloj I've put in solution with another approach here: https://github.com/Islandora-Devops/claw-playbook/pull/47

whikloj:
Closed in favour of https://github.com/Islandora-Devops/claw-playbook/pull/47

PR #51, opened 2018-01-24
Repo-Remi for php7 on CentOS

seth-shaw-unlv:
Addresses [Drupal 8 upcoming requirement for PHP7](https://groups.drupal.org/node/518200) (see also [this issue on the CLAW repo](https://github.com/Islandora-CLAW/CLAW/issues/754)).

Based on a collection of pulls (see [the related comment](https://github.com/Islandora-Devops/claw-playbook/pull/42#issuecomment-357432227)) on @DigitLib [fork](https://github.com/DigitLib/claw-playbook) who did it a bit cleaner than I originally had.


inventory/vagrant/group_vars/webserver/php-RedHat.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+php_date_timezone: "America/Toronto"
```
whikloj: This variable is the same as in `php-Debian.yml` so I'd pull it out into a separate variable file.

seth-shaw-unlv: Actually, I would be fine simply removing it. The default from geerlingguy is Chicago, but I'm in Las Vegas, so setting a default for Toronto (although consistent w/ Islandora's origin) doesn't really make sense to me. Might as well set it to GMT.

whikloj: @seth-shaw-unlv Good point @dannylamb @Natkeeran @jonathangreen what are your feelings on this, do we really need to set the datetime zone?

dannylamb: Sometimes things complain if it's not set.  I'd just make it configurable with a default of GMT.

whikloj: We discussed this at the CLAW call and the consensus is that as long as it is set (and it is set to [Chicago](https://github.com/geerlingguy/ansible-role-php/blob/462ebfaabf7ec41f609c35dae43c3d34cddf6760/defaults/main.yml#L63) as @seth-shaw-unlv said) then we can remove these

requirements.yml
```diff
@@ -1,3 +1,8 @@
+- src: geerlingguy.php-mysql
```
whikloj: @seth-shaw-unlv should this have a version to pin it to?

seth-shaw-unlv: We can. I will update the PR to pin the current version.

whikloj:
@seth-shaw-unlv looking at this one again, what loads the `php-{{ ansible_os_family }}.yml` file? Is that loaded from geerlingguy.php? I think my problem is that (and I can't explain why it did work before) the php-Debian.yml gets loaded each time and php-Redhat.yml gets ignored. So my centos/7 builds are failing on trying to load Ubuntu php packages.

seth-shaw-unlv:
I assume it is through geerlingguy (I'm not seeing it in ours). I think that piece, specifically, was taken from @DigitLib who might remember better.

DigitLib:
@whikloj and @seth-shaw-unlv it works when I test this soloution (it works a couple of times) . I tried now but it failed. Try to catch what is going on...

DigitLib:
@whikloj and @seth-shaw-unlv  renamed php-Debian.yml and php-RedHat.yml to Debian.yml and RedHat.yml and it passed. 
@whikloj Can you try to rename it and try?

seth-shaw-unlv:
Well, it appears to have passed for me, too.

DigitLib:
@seth-shaw-unlv Great! Now we wait will it pass @whikloj . It would be good if someone would try it on Ubuntu, so to PR if it work.

seth-shaw-unlv:
I tried it with both.

DigitLib:
Great!!

whikloj:
Okay that does work 👍 , my only question is... is that how it is supposed to work? Meaning is geerlingguy expecting us to provide OS specific variable this way?

DigitLib:
@whikloj I think so, because in [his  php role](https://github.com/geerlingguy/ansible-role-php) in ../tasks/main.yml he make this Variable setup.
- name: Include OS-specific variables.
include_vars: "{{ ansible_os_family }}.yml"

which is in ../var defined on that way. As you can see in that files he check packages etc...
I think it will recognize in our playbook this OS variables.

PR #55, opened 2018-02-01
Issue 791

seth-shaw-unlv:
Resolves [CLAW issue 791](https://github.com/Islandora-CLAW/CLAW/issues/791).


whikloj:
This seems to make sense to me, but I'm not a heavy Ansible user. I also tend to believe spurious internet claims (https://www.reddit.com/r/ansible/comments/5xh8u8/ansible_user_id_always_root/)

@dannylamb @Natkeeran @jonathangreen second opinions?

jonathangreen:
Looks like I'm the one that made the change... 

Trying to piece back together why I made the change. When your deploying on metal (or outside of vagrant really), perhaps you want to use another username... so you don't define `ansible_user` in your playbook variable then this returns an error because `ansible_user` isn't defined. 

Maybe we just require that people define `ansible_user` or we create a specific user for CLAW? I'm fine with reverting this for now, it shouldn't cause issues unless someone deploying isn't defining `ansible-user`.

PR #59, opened 2018-02-15
Fix for defining php variables for external role.

### LINKED ISSUE #801: claw-playbook - drupal error
claw-playbook gets deployed without error.  However, when I go to localhost:8000, the following error message gets thrown:

```
handle($request); $response->send(); $kernel->terminate($request, $response); 
```

```
commit 00fa2321fe8c2840ab3882d7f83e3d05ec3088f1
Author: Natkeeran <natkeeran@gmail.com>
Date:   Mon Feb 12 11:11:58 2018 -0500
```

whikloj:
Hopefully resolves https://github.com/Islandora-CLAW/CLAW/issues/801

To define the variables for use in geerlingguy.php to use different versions of `php_packages_extra` this seems to work as his role includes these [file names](https://github.com/geerlingguy/ansible-role-php/blob/master/tasks/main.yml#L4).

Perhaps the better way is to move this stuff to a `/vars` directory, but then I don't know how you set a variable in the inventory. Would we do that in the `bootstrap.yml`?

If there is a better way, please tell me. But this seems to work.

### Interested parties
@seth-shaw-unlv @DigitLib @Natkeeran @MarcusBarnes @jonathangreen


seth-shaw-unlv:
I think this is the way to go. 

We _could_ do it in vars + bootstrap.yml, but since geerlingguy's role already provides the logic this seems the better solution.

DigitLib:
I agree with @seth-shaw-unlv this solution work fine and follow a geerlingguy role-php.

seth-shaw-unlv:
Oddly enough, I just discovered that this doesn't actually work. Ansible doesn't complain and a CentOS  provision works, but when I spun up a Debian-based Islandora the apache php library wasn't there. I removed the Debian.yml file altogether and it actually worked then. To be honest, I'm not sure what is going on with the Debian php_packages_extra here.

whikloj:
Alright we can't hack geerlingguy's role, so we need to store these variables using a separate named variable file and include based on ansible_os_family at the start of the bootstrapping. I'm closing this.

whikloj:
Can someone try this out. It seems to work, but then I thought the last attempt worked too 🤣

DigitLib:
@whikloj for some reason geerligguy.php role didn't saw /var/RedHat.yml...
Note: when I try Ubunty php-Debian.yml installed but php-RedHat.yml do not, but RedHat.yml is installed from /group_var/webserver dir

Natkeeran:
@whikloj 

```
ERROR! the role 'geerlingguy.mysql' was not found in /home/nat/Desktop/clawtest2/claw-playbook/roles:/home/nat/Desktop/clawtest2/claw-playbook/roles/internal:/home/nat/Desktop/clawtest2/claw-playbook/roles/external:/home/nat/Desktop/clawtest2/claw-playbook

The error appears to have been in '/home/nat/Desktop/clawtest2/claw-playbook/database.yml': line 12, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
    - role: geerlingguy.mysql
      ^ here
```

whikloj:
@DigitLib / @Natkeeran thanks...I'll try burning my setup down and pulling fresh to see what is happening.

DigitLib:
@whikloj this /var files seem to work today. I try it on Manjaro Linux, yesterday it won't on Fedora 26... Tried Ubuntu and CentOS it works on both. Without added gather_facts: false...

whikloj:
Thanks @DigitLib/@Natkeeran, I am getting an error but not until building the Grok dependencies. That was on Centos, I'll tear it down and try again with Ubuntu

Natkeeran:
@whikloj You can ignore the previous error I posted.  I may not have installed the requirements! (was sleeping!).

This PR installs without error, but get the same php error related to missing library:
`handle($request); $response->send(); $kernel->terminate($request, $response); `

whikloj:
@Natkeeran @mjordan @DigitLib 
I ran the playbook with `-vvv` logging and it appeared that it wasn't actually finding the `php_extra_packages` variable. 

Moving it into the tasks and using `include_vars:` seems to work, and we can use this as a place to include overlapping variable definitions that need to be different based on OS version.

We could also use more specific files if necessary and the `with_first_found` argument to the `include_vars` command.
```
- name: Load a variable file based on the OS type, or a default if not found. Using free-form to specify the file.
  include_vars: "{{ item }}"
  with_first_found:
    - "{{ ansible_distribution }}.yml"
    - "{{ ansible_os_family }}.yml"
    - "default.yml"
```

seth-shaw-unlv:
Tested both Ubuntu and CentOS today. Both work for me now.

PR #60, opened 2018-02-26
Update requirements and group_vars to use updated tomcat role

whikloj:
Related to: https://github.com/Islandora-Devops/ansible-role-tomcat8/issues/2 and https://github.com/Islandora-Devops/ansible-role-tomcat8/pull/3

This updates the requirements to pull in the updated ansible-role-tomcat8 and defines some needed variables (as the defaults in ansible-role-fcrepo-syn are now incorrect).

### DIFF
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -23,7 +23,7 @@ tomcat8_java_opts:
   - -Dcantaloupe.config={{ cantaloupe_symlink }}/cantaloupe.properties
   - -Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true
 
-fcrepo_syn_tomcat_home: /opt/tomcat
+fcrepo_syn_tomcat_home: "{{ tomcat8_home }}"
 fcrepo_syn_default_public_key_path: "{{ fcrepo_syn_tomcat_home }}/conf/public.key"
 
 fcrepo_syn_sites:
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -23,7 +23,8 @@ tomcat8_java_opts:
   - -Dcantaloupe.config={{ cantaloupe_symlink }}/cantaloupe.properties
   - -Dorg.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH=true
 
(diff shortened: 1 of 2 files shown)


seth-shaw-unlv:
Thanks for doing the PR; I was just getting around to it.

whikloj:
@seth-shaw-unlv no worries, I figured you might be busy with...you know...work. 😄

seth-shaw-unlv:
Doesn't quite work. Running agains Ubuntu fails on the keymaster copy public key out because it wants to drop the key in /opt/tomcat/conf (which doesn't exist). At least the fcrepo_syn_tomcat_home variable should be moved to the OS-based vars files.

DigitLib:
I suggest that for both distos tomcat_home is /var/lib/tomcat8, this is path to tomcat in other roles, so we don't need to fix to new path on CentOS. I try this symlinked path on CentOS and work well with other roles.

seth-shaw-unlv:
I suppose that would be fine. I'm just leery of breaking with a distro's usual install locations for the sake of our own consistency. Really, this variable (and others like it) should be built off the tomcat8 variable.

whikloj:
I can just make `fcrepo_syn_tomcat_home: {{ tomcat8_home }}`  which should resolve the differences between OSes.

whikloj:
@seth-shaw-unlv @DigitLib I tested this with Ubuntu and it seems to work. Could you give it a spin and see if it works for you.

DigitLib:
@whikloj Tested on CentOS and work.
@seth-shaw-unlv I agree with you.

seth-shaw-unlv:
Tested ok here on both CentOS and Ubuntu.

whikloj:
@Islandora-CLAW/committers I believe this is ready

whikloj:
@seth-shaw-unlv sorry to do this to you, but I'm hoping once the other committers return someone will be able to burn through these quickly. So I merged https://github.com/Islandora-Devops/ansible-role-blazegraph/pull/5, added a tag and this now also updates that requirement.

whikloj:
@DiegoPino if you want to test this and merge it I am fine with that, there is also https://github.com/Islandora-Devops/claw-playbook/pull/59 which is causing problems that @Natkeeran. He might be testing that one though.

whikloj:
@dannylamb @jonathangreen @DiegoPino @Natkeeran can this get a review please.

DiegoPino:
@whikloj looks good. Merging.

PR #61, opened 2018-03-13
Use our drupal-project to pin drupal version to 8.4

### LINKED ISSUE #789: Followup with geerlingguy to see if drupal version can be specified
Currently we are planning to use https://github.com/Islandora-CLAW/drupal-project to specify a drupal version.  @jonathangreen suggested further research to see if it is possible to specify a drupal version in  https://github.com/geerlingguy/ansible-role-drupal.  This ticket is to follow up that issue.

whikloj:
Resolves: Islandora-CLAW/CLAW/issues/789

Copies the [`drupal-composer/drupal-project`](https://github.com/drupal-composer/drupal-project) and sets a version constraint on `drupal/core`. [Diff here](https://github.com/drupal-composer/drupal-project/compare/8.x...Islandora-CLAW:8.x-1.x).

Then we add a tag. This tag can be resolved from packagist with `islandora/drupal-project:tagName`, so we can update the `8.x-1.x` branch and make tags for following Drupal versions.

We will have to watch to ensure we don't get too far out of step with the base repo, as we will  have to merge changes into this repository to maintain our old tags.

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -14,7 +14,7 @@ drupal_composer_dependencies:
   - "drupal/facets:1.x-dev"
   - "islandora/carapace:dev-8.x-1.x"
   - "islandora/islandora_image:dev-8.x-1.x"
-drupal_composer_project_package: "drupal-composer/drupal-project:8.x-dev"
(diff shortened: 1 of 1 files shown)


whikloj:
@dannylamb @Natkeeran @jonathangreen

PR #62, opened 2018-04-06
Update grok dependency

whikloj:
Updates the Grok dependency to the new "works with Centos" version.

Please test with both Ubuntu and Centos. Though Centos will probably fail once you get past Grok 😄

### DIFF
--- a/crayfish.yml
+++ b/crayfish.yml
@@ -2,8 +2,13 @@
 
 - hosts: crayfish
   become: yes
+  vars:
+    php_enablerepo: "remi-php70"
+    php_packages_state: "latest"
 
   roles:
+    - name: geerlingguy.repo-remi
+      when: ansible_os_family == "RedHat"
     - geerlingguy.apache
     - geerlingguy.php
     - geerlingguy.git
--- a/requirements.yml
+++ b/requirements.yml
@@ -83,7 +83,7 @@
 
 - src: https://github.com/Islandora-Devops/ansible-role-grok
   name: Islandora-Devops.grok
-  version: 0.0.1
+  version: 0.0.2
 
 - src: https://github.com/Islandora-Devops/ansible-role-karaf
   name: Islandora-Devops.karaf


whikloj:
@dannylamb @jonathangreen @Natkeeran could someone look at this. 

Deploying claw-playbook on non-Ubuntu is still non-existant.
https://github.com/Islandora-CLAW/CLAW/issues/809#issuecomment-385464723

ajs6f:
Don't know how useful this is, but on a CentOS box I'm getting:
```
TASK [Islandora-Devops.grok : Cmake grok] **************************************************************************************************
Monday 30 April 2018  16:31:31 -0400 (0:00:00.977)       0:15:39.442 **********
fatal: [default]: FAILED! => {"changed": true, "cmd": ["cmake", "../"], "delta": "0:00:00.023255", "end": "2018-04-30 16:31:32.214446", "msg": "non-zero return code", "rc": 1, "start": "2018-04-30 16:31:32.191191", "stderr": "CMake Error at CMakeLists.txt:10 (cmake_minimum_required):\n  CMake 3.1 or higher is required.  You are running version 2.8.12.2", "stderr_lines": ["CMake Error at CMakeLists.txt:10 (cmake_minimum_required):", "  CMake 3.1 or higher is required.  You are running version 2.8.12.2"], "stdout": "-- Configuring incomplete, errors occurred!", "stdout_lines": ["-- Configuring incomplete, errors occurred!"]}
```
I will investigate further-- could be some chain of versions of dependencies going wrong?

ajs6f:
Okay, I was able to get past my fail above by installing `cmake3` and changing [this line](https://github.com/Islandora-Devops/ansible-role-grok/blob/master/tasks/install.yml#L38) to use `cmake3`, but now I'm exploding in flames in the next step (`make`) like [this](https://gist.github.com/ajs6f/cafde8c58ff6154bfbf2dc31e21c8a34). 

Starting to wonder if I can't just get away with disabling Grok...

ajs6f:
Hang on, I've done something wrong here-- I think I'm running parts of the Ansible play without this PR, and parts with. Or something like that.

ajs6f:
Okay, after further investigation, it is clear that Grok fails to build under any condition on CentOS7 (with something like [this](https://www.google.com/search?q=grok+error%3A+%E2%80%98chrono_literals%E2%80%99+is+not+a+namespace-name++using+namespace+std%3A%3Achrono_literals%3B&oq=grok+error%3A+%E2%80%98chrono_literals%E2%80%99+is+not+a+namespace-name++using+namespace+std%3A%3Achrono_literals%3B)), so this isn't a CLAW issue per se, it's upstream. I'm going to open a new ticket about this, because it is apparently CentoOS-specific (since other people are able to use this PR).

ajs6f:
The last comment is WRONG WRONG WRONG. See Islandora-CLAW/CLAW/issues/809 for the TRUTH.

seth-shaw-unlv:
Anyone else, @Islandora-Devops/committers,  want to review this before we merge?

seth-shaw-unlv:
Covered by #68.

PR #63, opened 2018-04-06
Rename Redhat.yml to RedHat.yml

DigitLib:
This fix rename the file so it can be readed as OS-specific variable.


whikloj:
Hi @DigitLib.

Is this causing an error and if so could you explain it here so we can test it.

DigitLib:
@whikloj Yes. It can't read file, it don't recognise it as Redhat when rename it was readed

dannylamb:
I see we've got some dynamic naming going on there.  Nice catch @DigitLib.  Thanks for the PR!

PR #64, opened 2018-04-07
claw_db undefined error fixed

DigitLib:
fatal: [default]: FAILED! => {"msg": "The conditional check 'crayfish_db == 'mysql'' failed. The error was: error while evaluating conditional (crayfish_db == 'mysql'): 'claw_db' is undefined\n\nThe error appears to have been in '/home/andr/CLAW-latest/claw-playbook/roles/external/Islandora-Devops.crayfish/tasks/db-mysql.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Create Gemini DB (mysql)\n  ^ here\n"}

When changed with crayfish_db_name installed on CentOS. Please check it on Debian.

### DIFF
--- a/inventory/vagrant/group_vars/crayfish.yml
+++ b/inventory/vagrant/group_vars/crayfish.yml
@@ -1,6 +1,6 @@
 ---
 
-crayfish_db: "{{ claw_db }}"
+crayfish_db: "{{ crayfish_db_name }}"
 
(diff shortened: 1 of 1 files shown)


DigitLib:
My error when working with clayfish, exluded database.yml accidently.

PR #65, opened 2018-04-18
Increase the Vagrant version to allow 2.0.3

ysuarez:
Islandora-vagrant forced 2.0.3 for so we should update too.


ysuarez:
@dannylamb Do I need to do anything else to have this commit merged?

whikloj:
@ysuarez You're good. I'll try it out and let you know or merge it.

PR #70, opened 2018-07-10
Permit current Vagrant version.

rosiel:
The playbook currently is known to work on versions of vagrant up to 2.1.2 (the current). But it involves manually editing the Vagrantfile to spin it up unless you have an old version of Vagrant.

I've done this in a way that we don't specify a "max" allowed version, because I don't think it's necessary until something is known to break. This way people who are just starting with vagrant don't have to re-install an older version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


dannylamb:
Whoa, just wound up doing this myself recently.  @rosiel++

PR #71, opened 2018-07-24
Workaround for content_translation columns not getting added when imp…

dannylamb:
…orting config

Resolves Islandora-CLAW/CLAW#873

After pulling down this code and running `vagrant up` to get a fresh box, the install should yield a functional environment.  I tested by adding an image and verifying that derivatives were generated and that they could be pulled out of Fedora.

Note, I have to purposefully run a command that fails during the post-install commands.  You'll get a big wall of red text from the fail, but then should see `ignoring...` underneath it and the play continue on.  That's to be expected.  We can stop doing it that way once https://www.drupal.org/node/2599228 lands in Drupal 8.7.

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -38,7 +38,6 @@ drupal_enable_modules:
   - search_api_solr_defaults
   - facets
   - islandora_core_feature
-  - islandora_demo_feature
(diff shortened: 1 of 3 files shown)


DigitLib:
Passed

ajs6f:
The "wall of red" is from `[Install demo module (fail ok)]`, right?

ajs6f:
Woo-hoo! Successful install!

PR #79, opened 2018-10-03
Installation for testing Islandora-CLAW/CLAW#888

dannylamb:
Use this to test all the PRs for Islandora-CLAW/CLAW#888

Resolves https://github.com/Islandora-CLAW/CLAW/issues/921

After you vagrant up, you should see several important changes.
- The `resource_types` vocabulary is filled with DCMI Type entries
screenshot from 2018-10-03 10-54-54
- The form is divided into fieldsets using the `field_group` module
screenshot from 2018-10-03 10-55-23
- You can use `content_browser` for `field_member_of` and `field_media_of` to choose from existing nodes
screenshot from 2018-10-03 10-58-22
- You can also use inline entity form to create new nodes and subjects from within the parent form
screenshot from 2018-10-03 10-59-06
- The `model` field has a drop-down instead of an autocomplete
screenshot from 2018-10-03 11-00-06
- There's `access control` and `display hints` fields, as well.
- Access control uses `field_permission` so that it's private to admins and the person who created the content
screenshot from 2018-10-03 11-01-50

Starting to look more full featured for sure!

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -14,9 +14,7 @@ drupal_composer_dependencies:
   - "drupal/matomo:^1.7"
   - "islandora/carapace:dev-8.x-1.x"
   - "islandora/openseadragon:dev-8.x-1.x"
-  - "islandora/islandora_image:dev-8.x-1.x"
-  - "islandora/islandora_demo:dev-8.x-1.x"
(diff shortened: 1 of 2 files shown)


dannylamb:
Re-testing with 28b3e6561f0b9fd94be83d7a7af49052f0688d29 to make sure it still works after all the changes I'm making to `islandora_demo` to get tests working.  May have to update https://github.com/Islandora-CLAW/CLAW/blob/master/.scripts/travis_setup_drupal.sh as well.  So hang tight.

PR #80, opened 2018-10-11
Claw 948

ppound:
create simpletest directories, chown feature modules so they are owned by www-data and add ubuntu or vagrant user to appropriate group


dannylamb:
`vagrant up`

PR #89, opened 2019-01-08
Refactor blueprint

dannylamb:
Resolves https://github.com/Islandora-CLAW/CLAW/issues/957
Resolves https://github.com/Islandora-CLAW/CLAW/issues/974
Resolves https://github.com/Islandora-CLAW/CLAW/issues/982

Use this PR to test those issues.  This points the installer to my various `refactor-blueprint` branches across the stack to enable video derivatives.

## Testing Instructions

- Pull in this PR
- Delete your roles/external directory
- `vagrant up`

Once it's up, make a `Repository Item` and give it a Video media tagged as Original File.  Wait a bit and eventually you'll get service file derivative created :rocket: 

I've noticed that about half the time, there's issues with `fcrepo-indexing-triplestore` and `fcrepo-api-x` from deployment that will clog up the queue for a bit.  I think the issue has been there since we switched to Fedora 5, but I can't be sure.  It's intermittent, so if it's slow for you that's probably what's happening.  But be patient, it should sort itself out eventually and give you a derivative.

### DIFF
--- a/inventory/vagrant/group_vars/karaf.yml
+++ b/inventory/vagrant/group_vars/karaf.yml
@@ -1,30 +1,6 @@
 ---
 
-alpaca_settings:
-  - pid: ca.islandora.alpaca.http.client
-    settings:
-      token.value: islandora
-  - pid: ca.islandora.alpaca.connector.broadcast
-    settings:
-      input.stream: activemq:queue:islandora-connector-broadcast
-  - pid: ca.islandora.alpaca.indexing.triplestore
-    settings:
-      error.maxRedeliveries: 10
-      index.stream: activemq:queue:islandora-indexing-triplestore-index
-      delete.stream: activemq:queue:islandora-indexing-triplestore-delete
-      triplestore.baseUrl: http://localhost:8080/bigdata/namespace/islandora/sparql
-  - pid: ca.islandora.alpaca.indexing.fcrepo
-    settings:
-      error.maxRedeliveries: 5
-      node.stream: activemq:queue:islandora-indexing-fcrepo-content
-      node.delete.stream: activemq:queue:islandora-indexing-fcrepo-delete
-      media.stream: activemq:queue:islandora-indexing-fcrepo-media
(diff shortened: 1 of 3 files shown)


Natkeeran:
@dannylamb 

Works as advertised.

dannylamb:
@Natkeeran Ok, there's a lot of OSGi magic but here goes:

I renamed `islandora-connector-houdini` to `islandora-connector-derivative` and removed its blueprint and cfg files.  This kinda turns it into a library.  The routes available for other programs to use, but it doesn't actually start a listener.  I took the config and put it into the ansible role so other people can override it.  The role now templates out a blueprint file for each listener we want to spin up using that config.  When ansible puts the blueprint file in `/opt/karaf/deploy`, it applies the config to the camel route and starts listening.  We're essentially dynamically generating a program by deploying a configuration file.  Instead of code re-use, it's like "whole program re-use".

So in theory, if you made another microservice and configured an appropriate Drupal action to publish the message on a queue, then reliably hooking the two together boils down to another chunk of yml config here: https://github.com/Islandora-Devops/ansible-role-alpaca/pull/2/files#diff-7eeda618087b49ae876084ab6c73fdbbR44

Natkeeran:
@dannylamb Thank you for the info.  Don't quite follow it.  But, will come back to it when needed.  

Wondering if we should create a TN derivative file for video.

dannylamb:
@whikloj That should work, but we won't know until we merge the other PRs and test this.  We'll have to slice new versions for both Alpaca and its ansible role after merging.

whikloj:
I have merged all 4 PRs, added tags to ansible-role-alpaca and cut and released the Alpaca 0.7.0 release.

Wondering if [this](https://github.com/Islandora-Devops/ansible-role-alpaca/blob/master/defaults/main.yml#L9) should be set to an actual release and updated regularly. Probably not there, but in this repo?

dannylamb:
@whikloj YUP, we'll want to pin that down to actual versions for sure.  Just going to link this to https://github.com/Islandora-CLAW/CLAW/issues/950 for now.

PR #94, opened 2019-02-08
Update requirements.yml

dannylamb:
Bumping up crayfish role version number for Islandora-CLAW/CLAW#930


dannylamb:
@Islandora-Devops/committers Bump... this'll get us a PDF viewer and audio derivatives with the installer.

PR #96, opened 2019-03-01
Update for drupal core update

whikloj:
Related to https://github.com/Islandora-CLAW/drupal-project/pull/26

## DO NOT MERGE THIS

This has various updates to allow the new version of drush that seems to come with updating our drupal-project to work.

But it also points to a PR branch of work.

Use this for testing, pull it down and bring it up and everything should work.

If this is good:
1. merge the above [drupal-project PR](https://github.com/Islandora-CLAW/drupal-project/pull/26
)
1. add a new tag to the drupal-project repo.
1. Then I will update this PR with that tag and this can be re-tested and merged normally.

@Islandora-Devops/committers @seth-shaw-unlv

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -17,7 +17,8 @@ drupal_composer_dependencies:
   - "islandora/carapace:dev-8.x-1.x"
   - "islandora/openseadragon:dev-8.x-1.x"
   - "islandora/islandora_demo:dev-8.x-1.x"
-drupal_composer_project_package: "islandora/drupal-project:8.6"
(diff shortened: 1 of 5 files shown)


seth-shaw-unlv (approved):
Works.

seth-shaw-unlv:
@whikloj do want to just revert the inventory config and we can go ahead and merge the rest (which works even with the current drupal-project) so it is ready when the [drupal-project PR](/Islandora-CLAW/drupal-project/pull/26) is ready to go? We can update the inventory separately after the new drupal-project version gets cut.

seth-shaw-unlv:
Testing again right now, just to be safe.

seth-shaw-unlv:
:shipit:

PR #98, opened 2019-03-22
Update requirements to pull new Crayfish

whikloj:
This tag pulls the updated Crayfish with the Recast service


whikloj:
@Islandora-Devops/committers simple update here.

whikloj:
Forget this, I see that the role doesn't include config.yaml for Recast.

PR #103, opened 2019-04-29
Always run this pre-task, even when using tags

whikloj:
This has no impact unless you try to use `ansible-playbook` with the `--tags` flag, then this is skipped and any use of it will fail.

@Islandora-Devops/committers


dannylamb (approved):
@whikloj++ I always end up defining this manually every time i try to run just a tag.

PR #113, opened 2019-06-06
Removes vm's audio

DonRichards:
VM is triggering an action to take over control for the audio (mic & speakers). While this can be expected in a "Windows 10" VM or something, it should probably not exist in a headless VM. The base image has it so the best option at this point is to tell vagrant not to enable this in the first place. 

If you have headphones on you'll hear the speaker click or sound different when the system boots. I'm getting a notice from some software that my mic is being accessed by virtualbox whenever the VM is running.


DonRichards:
This PR is trivial at best. Most won't notice the audio interruption and it's difficult to check if it works in most environments. It shouldn't have any negative effects. Audio files and the sound from videos will still be sent to the browser and everything. Te audio feature is really meant for GUI VMs and not headless.

PR #115, opened 2019-06-19
Reset External Roles

DonRichards:
Simplifies the process of resetting the external roles.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -64,6 +64,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision "reset_roles",
     type: "shell",
     preserve_order: true,
-    inline: "cd /vagrant/roles && rm -rf external/"
+    inline: "rm -rf /vagrant/roles/external"
   end
 end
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -58,5 +59,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     }
     ansible.extra_vars = { "islandora_distro" => $vagrantBox }
   end
-
+  unless  $reset_roles.eql? "FALSE"
+    # Fires last to modify one last change.
```
whikloj: If this fires last, wouldn't it delete after the ansible-galaxy command? Because that is the one that populates the roles/external with the external roles.

DonRichards: Yes, and no. It shouldn't fire at all unless you're explicitly trying to run it before you rebuild. Assuming you have already ran vagrant up and need to wipe out your external roles before rebuilding, you could run the `ISLANDORA_VAGRANT_REST_ROLES='TRUE' vagrant reload` then destroy and up. I know we could easily just delete the folder ourselves but this automated the process for people that are new at it.

DonRichards: The no part was, No it shouldn't ever fire unless you specify it in the provisions part. And if you run it as suggested it should be the only provision to fire regardless.

DonRichards:
The hole PR may or may not be helpful. Maybe a document update or a ansible task is more appropriate?

DonRichards:
Is there any interest in this or should we close this?

PR #141, opened 2019-09-18
retries and checks

DonRichards:
**GitHub Issue**: Not sure

Adds tags, checks and retries

# What's new?
tags, checks and retries

# How should this be tested?
change the URL of the solr zip to something that will fail to see if it actually fails.

# Additional Notes:
N/A

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/solr.yml
+++ b/solr.yml
@@ -58,3 +58,5 @@
     - name: Fetch a Solr query to warm up the application
       command: wget -q -O- http://localhost:8983/solr/ISLANDORA/select?q=warm
       tags: solr
+      args:
+          warn: no
--- a/solr.yml
+++ b/solr.yml
@@ -2,6 +2,7 @@
 
 - hosts: solr
   become: yes
+  debugger: on_failed
 
   roles:
     - role: geerlingguy.java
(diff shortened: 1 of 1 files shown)


dannylamb:
Hey nice @DonRichards ! I take it this is a pattern we can use all over the place?

DonRichards:
@dannylamb Mostly, I'm trying to see what can be done with "unarchive". It looks highly limited but switching it to 2 functions (1 to download and the other to unarchive) seems to be the only solution I'm finding to make sure a file downloaded before moving on.

DonRichards:
```yaml
register: solr_config_download_and_extract
until: solr_config_download_and_extract is not failed
retries: 5
tags: solr
```

But the format should work. It requires a unique `registered` label
Everything has a '**failed**' (either true or false) status so `Until` requires the register label followed by `is not failed` 
retries default to 3 (although in the PR is 5, I'm toying with 15 with success almost 100% of the time). 
Tags should stay consistent to help with debugging

DonRichards:
Argggg..... Pushed to the wrong branch.

seth-shaw-unlv:
@DonRichards, rather than squashing the warning, would it make sense to use the [uri module](https://docs.ansible.com/ansible/latest/modules/uri_module.html) like it recommends?

DonRichards:
Sounds good, I'll make the change.

seth-shaw-unlv:
Sweet. I'm building a box right now but I'll give this a spin after that one completes.

DonRichards:
The delay looks long. I'm gonna fix that.

DonRichards:
Sorry about that. 30 seconds seemed too long between retries. I think 20 retries of a 60 second period should work.

DonRichards:
Ready to test.

seth-shaw-unlv:
vagrant 🚀

seth-shaw-unlv:
Looks good to me. :shipit:

PR #142, opened 2019-09-25
Performance and error message improvements

DonRichards:
**GitHub Issue**: N/A

Improves the build time and limits the error messages to avoid confusion.

# What's new?
**fact_caching**: Gather the facts once in a local file so that it only has to it lookup and not call it through secure shell.
**fact_caching_connection**: Facts cache location.
**poll_interval**: The time Ansible waits to poll for results.
**internal_poll_interval** = The time Ansible control host waits to poll for results.
**command_warnings** = Hides command warnings.
**deprecation_warnings**: Hides deprecation warnings.
**pipelining**: Enabling pipelining reduces the number of SSH operations required to execute a module on the remote server, by executing many ansible modules without actual file transfer.

# How should this be tested?
Start the VM without provisions first (to isolate the performance to the playbook as much as possible). Then using "time" to calculate the total time at the end to help, this is redundant because of the report Ansible gives but I do it anyways.
```shell 
$ vagrant destroy
$ vagrant up --no-provision
$ time vagrant provision
```
Keep track of _time_. Run the destroy/up/provision commands again with this PR to compare and test. 

You should see all of the these messages 
Before

turn into this

After

You should see theoretically an improved build time. This is difficult to measure due to the nature of the servers the playbook(s) are pulling from and the performance variations they incur. 

# Additional Notes:
There are several other things that need to be done to improve performance like pre-downloading all tars locally, grouping and setting async on those tasks that are not a dependency for another task to start. This should be a good start but will probably show a larger impact as the other issues are improved upon.

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,4 +1,16 @@
+# config file for ansible
+# ===============================================
+
 [defaults]
 host_key_checking = false
 roles_path = roles/internal:roles/external
 callback_whitelist = profile_tasks
+fact_caching = jsonfile
+fact_caching_connection = /tmp/.ansible_fact_cache
+poll_interval = 5
+internal_poll_interval = 0.001
+command_warnings = False
+deprecation_warnings = False
+
+[ssh_connection]
+pipelining = True


DonRichards:
Gonna need to rebase.

DonRichards:
OK, ready for testing.

dannylamb:
@DonRichards Testing now :rocket:

dannylamb:
I've spun up several boxes with this.  It's hard to pin down any improvements b/c my internet dl speed waxes and wanes with the phases of the moon, but I can say it _feels_ snappier and definitely doesn't break anything.  Less verbosity also makes the installer somewhat more approachable.

PR #146, opened 2019-12-04
specify alpaca_karaf_repos to include 1.0.2 of alpaca

elizoller:
Tells the ansible-role-alpaca to use version 1.0.2 of Alpaca

# What's new?
A change in the karaf.yml for inventory/vagrant/group_vars that overrides the defaults for ansible_karaf_repos in the ansible-role-alpaca

# How should this be tested?
* check out PR branch
* run clean install
* make a repository item
* check to make sure that repository item made it to fedora

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/inventory/vagrant/group_vars/karaf.yml
+++ b/inventory/vagrant/group_vars/karaf.yml
@@ -4,6 +4,10 @@
  alpaca_from_source: yes
  alpaca_version: dev
  alpaca_clone_directory: /opt/alpaca
+ alpaca_karaf_repos:
+  - mvn:org.apache.camel.karaf/apache-camel/2.20.4/xml/features
+  - mvn:org.apache.activemq/activemq-karaf/5.15.0/xml/features
+  - mvn:ca.islandora.alpaca/islandora-karaf/1.0.2/xml/features
 
  triplestore_namespace: islandora
  alpaca_triplestore_base_url: "http://{{ hostvars[groups['tomcat'][0]].ansible_host  }}:8080/bigdata"


dannylamb:
That should work @elizoller, but I can't seem to get a server up for reasons totally unrelated to this pull.  Ever since I updated to Ubuntu 18.04 everything's a mess....

inventory/vagrant/group_vars/karaf.yml
```diff
@@ -4,6 +4,10 @@
  alpaca_from_source: yes
  alpaca_version: dev
  alpaca_clone_directory: /opt/alpaca
+ alpaca_karaf_repos:
+  - mvn:org.apache.camel.karaf/apache-camel/2.20.4/xml/features
+  - mvn:org.apache.activemq/activemq-karaf/5.15.0/xml/features
+  - mvn:ca.islandora.alpaca/islandora-karaf/1.0.2/xml/features
```
whikloj: There is no `1.0.2` version of Alpaca.
https://search.maven.org/artifact/ca.islandora.alpaca/islandora-karaf

elizoller: I was going off this: https://github.com/Islandora/Alpaca/blob/dev/gradle.properties

dannylamb: We're building it from source now in dev.  This is to make it work that way.  We could also just slice a 1.0.2 release and do it that way.

elizoller: For what it's worth, I did test a build with this PR and successfully built Alpaca from dev and installed the features. Without this PR, even with telling it to build from dev, it only installs the 1.0.1 version of Alpaca so it is missing the versioning components and breaks the sending of content to Fedora.

whikloj: That should probably be looked at in the ansible-role-alpaca then, it must not be working as expected. Probably some ansible version issue.

elizoller: Am I reading this wrong? It looks like to me that the ansible-role-alpaca includes the compile.yml task when the `ansible_from_source` is true but in compile.yml (https://github.com/Islandora-Devops/ansible-role-alpaca/blob/master/tasks/compile.yml#L26) it doesn't appear to be doing a gradle `clean build install` but rather just an `install`. My knowledge of gradle is limited but I'm guessing we at least need the `build` part if we're building from dev rather than an existing version?

elizoller: Am I reading this wrong? It looks like to me that the ansible-role-alpaca includes the compile.yml task when the ansible_from_source is true but in compile.yml (https://github.com/Islandora-Devops/ansible-role-alpaca/blob/master/tasks/compile.yml#L26) it doesn't appear to be doing a gradle clean build install but rather just an install. My knowledge of gradle is limited but I'm guessing we at least need the build part if we're building from dev rather than an existing version?

whikloj: Clean is only needed if you built if before, so as we (should) only build it once the clean is redundant. 

But after some digging I see why you need this. We will need to watch and ensure that as the version of alpaca in `dev` moves up so does this variable.

I'm going to pull this in and rebuild my vagrant.

dannylamb: and unfortunately `alpaca_version` can be a branch name, so it's not like we can put it in the repo url to keep things in sync :(

whikloj:
Just remembered I need to test build with Centos before merging

PR #155, opened 2020-01-24
added microservice variables (the re-do)

ajstanley:

Add some variables to configure microservices

# What's new?
 Put in variables to configure fits and ocr microservices

# How should this be tested?

 Testing will come later when we can use these variables to configure karaf blueprints

# Interested parties
@dannylamb @whikloj


whikloj:
@ajstanley to confirm, there are no roles that use these variables yet?

ajstanley:
@whikloj -the Alpaca role hardcodes ` alpaca_fits_base_url: http://localhost:8000/crayfits` 
(so not changeable until it has been downloaded)
and 
 `alpaca_hypercube_base_url` isn't used at all yet
.

PR #156, opened 2020-01-29
[ENHANCEMENT] Make mem and cpu dynamic

DonRichards:
**GitHub Issue**: https://github.com/Islandora/documentation/issues/1413

1. Assess how much ram is available.
1. If "ISLANDORA_VAGRANT_MEMORY" is set use it. Otherwise If "ISLANDORA_VAGRANT_MEMORY" is blank use whichever is larger, 1/2 the RAM or 4GB.
1. If "ISLANDORA_VAGRANT_CPUS" is set use it. Otherwise if "ISLANDORA_VAGRANT_CPUS" is blank and there are no other VMS running set the CPU to 1/2 of the available CPUs, otherwise default to 1 core.

# What's new?
It looks up if VMS are running so that it doesn't over allocate the VM resources.

# How should this be tested?
* Test on a MAC, Linux, and windows machine to verify the test commands don't fail on a specific OS.
* Test if the 2 variable overrides work "ISLANDORA_VAGRANT_MEMORY" & "ISLANDORA_VAGRANT_CPUS"
* Test that the defaults are triggered when different VM is already running. This avoids issues will allocating too much RAM/CPUs collectively. It was brought up on the CLAW call as a check that would be useful.

```shell
# How many Cores
$ lscpu

# How much memory
$ free -h | gawk  '/Mem:/{print $2}'
```
# Additional Notes:
N/A

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,12 +9,15 @@ host = RbConfig::CONFIG['host_os']
 
 # List memory
 if host =~ /darwin/
+  $cpus = `sysctl -n hw.ncpu`.to_i
   # conver to MB
   $mem = `sysctl -n hw.memsize`.to_i / 1024 / 1024
 elsif host =~ /linux/
+  $cpus = `nproc`.to_i
   $mem = `grep 'MemTotal' /proc/meminfo | sed -e 's/MemTotal://' -e 's/ kB//'`.to_i / 1024
 else
   # else windows commands
+  $cpus = `wmic cpu get NumberOfCores`.split("\n")[2].to_i
   $mem = `wmic OS get TotalVisibleMemorySize`.split("\n")[2].to_i / 1024
 end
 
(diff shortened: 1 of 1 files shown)


DonRichards:
The CPU detect is off. Working on it.

DonRichards:
Ready for testing. @dannylamb @Islandora-Devops/committers

DonRichards:
Screen Shot 2020-01-30 at 10 38 05 AM

From the host machine I ran this. Not sure what to suggest for linux or windows to make the comparison but it should be easy enough.
```shell
$ sysctl hw.ncpu
hw.ncpu: 8

$ system_profiler SPHardwareDataType | grep "  Memory:"
Memory: 64 GB
```

DonRichards:
@Islandora-Devops/committers Is there any interest in having this merged?

PR #157, opened 2020-02-06
update search_api_solr to 3.8

elizoller:

Updates search_api_solr to version 3.8

# What's new?
* Updates [search_api_solr](https://www.drupal.org/project/search_api_solr) to version 3.8
* Per the [install.md](https://git.drupalcode.org/project/search_api_solr/blob/8.x-3.x/INSTALL.md) - adds a patch for Drupal Core
* Per the [install.md](https://git.drupalcode.org/project/search_api_solr/blob/8.x-3.x/INSTALL.md) - uses drush command to generate the default solr config files
* Reorders the playbook.yml so that solr.yml runs after webserver.yml because the solr config setup requires the output of the drush command

# How should this be tested?

* check out this PR, vagrant up
* check to see that you got a running solr instance with a core called ISLANDORA
* check to see that you have a default solr server in the drupal search api config (http://localhost:8000/admin/config/search/search-api/)
* check that http://localhost:8000/admin/config/search/search-api/server/default_solr_server says the schema is set to drupal-8.3.8-solr-7.x
* check that you can create an object and that it is indexed in solr and searchable

# Additional Notes:
I was concerned about conflating the solr host and the webserver host so that is why I implemented the "Unarchive solr config" as a request to localhost instead of directly from a file path like `/var/www/...` 

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/roles/internal/webserver-app/tasks/solr.yml
+++ b/roles/internal/webserver-app/tasks/solr.yml
@@ -7,7 +7,7 @@
   composer:
     command: require
     arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' drupal/search_api_solr
-    working_dir: "/var/www/html/drupal"
+    working_dir: "{{ drupal_composer_install_dir }}"
 
 - name: Set default solr server host from hostvars
   command: "{{ drush_path }} --root {{ drupal_core_path }} -y config-set search_api.server.default_solr_server backend_config.connector_config.host {{ hostvars[groups['solr'][0]].ansible_host  }}"
--- a/roles/internal/webserver-app/tasks/solr.yml
+++ b/roles/internal/webserver-app/tasks/solr.yml
@@ -6,7 +6,7 @@
 - name: Add a patch for symfony-event-dispatcher per the INSTALL.md docs on search_api_solr:3.8
   composer:
     command: require
-    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' solarium/solarium:^5.1.4 drupal/search_api_solr
+    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' solarium/solarium:^5.1 drupal/search_api_solr
     working_dir: "{{ drupal_composer_install_dir }}"
 
 - name: Set default solr server host from hostvars
--- a/roles/internal/webserver-app/tasks/solr.yml
+++ b/roles/internal/webserver-app/tasks/solr.yml
@@ -3,6 +3,11 @@
 
 # Only tasks related to configuring search_api_solr are put here
 # Solr server configuration should go into the solr playbook: https://github.com/Islandora-Devops/claw-playbook/blob/dev/solr.yml
+- name: Add a patch for symfony-event-dispatcher per the INSTALL.md docs on search_api_solr:3.8
+  composer:
+    command: require
+    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' drupal/search_api_solr
+    working_dir: "/var/www/html/drupal"
 
 - name: Set default solr server host from hostvars
   command: "{{ drush_path }} --root {{ drupal_core_path }} -y config-set search_api.server.default_solr_server backend_config.connector_config.host {{ hostvars[groups['solr'][0]].ansible_host  }}"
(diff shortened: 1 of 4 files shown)


roles/internal/webserver-app/tasks/solr.yml
```diff
@@ -3,6 +3,11 @@
 
 # Only tasks related to configuring search_api_solr are put here
 # Solr server configuration should go into the solr playbook: https://github.com/Islandora-Devops/claw-playbook/blob/dev/solr.yml
+- name: Add a patch for symfony-event-dispatcher per the INSTALL.md docs on search_api_solr:3.8
+  composer:
+    command: require
+    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' drupal/search_api_solr
+    working_dir: "/var/www/html/drupal"
```
whikloj: Can be `{{ drupal_composer_install_dir }}`
https://github.com/Islandora-Devops/islandora-playbook/blob/dev/inventory/vagrant/group_vars/webserver/drupal.yml#L4

elizoller: Thanks, I'll fix that

elizoller: fixed with 
036287f

seth-shaw-unlv:
Results in a warning: "Solarium 5.0.2 is too old and has known issues. " 

To alleviate this we can update the composer task to be: 
```yml
- name: Add a patch for symfony-event-dispatcher per the INSTALL.md docs on search_api_solr:3.8
  composer:
    command: require
    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' solarium/solarium:^5.1.4 drupal/search_api_solr
    working_dir: "{{ drupal_composer_install_dir }}"
```

I'm building a new instance with this revised task right now.

elizoller:
awesome, thanks @seth-shaw-unlv

seth-shaw-unlv:
So, this works!
Screen Shot 2020-03-12 at 9 03 55 AM

HOWEVER, it breaks the existing islandora_search configs. With the updated search_api_solr module they appear to assume solr will be accessible @ '/solr/' so the [linked config line](https://github.com/Islandora/islandora_defaults/blob/8.x-1.x/modules/islandora_search/config/optional/search_api.server.default_solr_server.yml#L19) turns `http://127.0.0.1:8983/solr/#/ISLANDORA` into `http://127.0.0.1:8983/solr/solr/#/ISLANDORA`. I don't know why they did that, but there it is.

So, @elizoller , if you can add that solarium bit to the composer command in this PR we can merge it. I'll create a new PR on islandora_defaults to purge the path config.

roles/internal/webserver-app/tasks/solr.yml
```diff
@@ -3,6 +3,11 @@
 
 # Only tasks related to configuring search_api_solr are put here
 # Solr server configuration should go into the solr playbook: https://github.com/Islandora-Devops/claw-playbook/blob/dev/solr.yml
+- name: Add a patch for symfony-event-dispatcher per the INSTALL.md docs on search_api_solr:3.8
+  composer:
+    command: require
+    arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' drupal/search_api_solr
```
seth-shaw-unlv: Replace with: `arguments: symfony/event-dispatcher:'4.3.4 as 3.4.99' solarium/solarium:^5.1.4 drupal/search_api_solr`

elizoller: ok, i'll update this

seth-shaw-unlv: Actually, there is a 5.1.6 already, so maybe we should use `^5.1`. I think composer should run with that...

elizoller: ok, spinning that up now

seth-shaw-unlv:
Spinning up another instance to double check before merge.

elizoller:
This build completed successfully for me and no more solarium warning

seth-shaw-unlv:
I had forgotten to pull in the Grok playbook fix for CentOS so my build died. 🤦‍♂️ It is building again and should be done in a little while.

PR #158, opened 2020-02-06
[ENHANCEMENT] Dynamic drupal trusted hosts setting.

cafuego:

This change allows direct access to Drupal via the virtual machine IP (or hostname) without a change to the Drupal `$settings['trusted_host_patterns']` variable.

# What's new?

Allows easier access to Islandora on the VM. However, this is not necessarily a safe change for a production installation.

# How should this be tested?

Access Islandora via a non-localhost address after VM creation.

# Interested parties

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -58,6 +58,7 @@ drupal_enable_modules:
 drupal_trusted_hosts:
   - ^localhost$
   - "{{ hostvars[groups['webserver'][0]].ansible_host }}"
+  - "^' . preg_quote($_SERVER['SERVER_NAME']) . '$"
 drupal_trusted_hosts_file: "{{ drupal_core_path }}/sites/default/settings.php"
 drupal_public_filesystem: "{{ drupal_core_path }}/sites/default/files"
 drupal_external_libraries_directory: "{{ drupal_core_path }}/libraries"


inventory/vagrant/group_vars/webserver/drupal.yml
```diff
@@ -58,6 +58,7 @@ drupal_enable_modules:
 drupal_trusted_hosts:
   - ^localhost$
   - "{{ hostvars[groups['webserver'][0]].ansible_host }}"
+  - "^' . preg_quote($_SERVER['SERVER_NAME']) . '$"
```
whikloj: This seems like a PHP command...does that work within the ansible framework which is (I think) Ruby?

dannylamb: I think it works because settings.php is a php file?  I haven't tried it yet to find out.

whikloj: Ohhhhhhh 🤦‍♂

cafuego: Yes, running ansible with this entry added to the variables will add `'^' . preg_quote($_SERVER['SERVER_NAME']) . '$'` to the trusted host patterns. That means any name (or IP) you use to access the Islandora instance in the VM will automagically be trusted, so you don't need to explicitly add it.

PR #160, opened 2020-02-07
Modifications for speed and re-runnability.

jhujasonw:
**GitHub Issue**:N/A

* Other Relevant Links N/A

There are 2 patches included.  One to the ansible.cfg file to speed up package installs, when enabled(it's commented out by default.) And one to re-set the perms on the default/ folder in drupal/sites/ so that the playbook is re-runnable. It keeps composer from dying badly.

# What's new?

* Added squash_actions  to ansible.cfg but commented out by default.
* Added some tasks to the bootstrap.yml to change perms on the default/ folder 
* Does this change require documentation to be updated? Perhaps only to mention that the squash_actions is there and could speed things up.
* Does this change add any new dependencies? no
 (i.e. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? the ansible.cfg option might speed things up a bit.  The changes to bootstrap.yml should not impact existing execution.

# How should this be tested?
Prior to this PR, if you tried to re-run the playbook, it would fail on composer errors.  The bootstrap.yml modifications fix this.  It is not clear what re-running this on a running system will do though.  More testing by someone with a bit more Islandora knowledge would be good before saying it's 100% safe to re-run on a working system.

# Additional Notes:

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -12,5 +12,8 @@ internal_poll_interval = 0.001
 command_warnings = False
 deprecation_warnings = False
 
+# uncomment to speed up the install a bit by squashing package installations
+#squash_actions = apk,apt,dnf,homebrew,pacman,package,pkgng,yum,zypper
+
 [ssh_connection]
 pipelining = True
--- a/bootstrap.yml
+++ b/bootstrap.yml
@@ -65,3 +65,21 @@
     - name: change selinux enforcement to permissive
       command: setenforce permissive
       when: ansible_os_family == "RedHat" and disable_selinux and selinux_permissive == "enforcing"
+
+    # Check if we may be re-running
+    - name: Check for possible re-run
+      stat:
+        path: "{{ drupal_composer_install_dir }}/web/sites/default/"
+      register: defaultdir_exists
+   
+    # if we are re-running, change the perms on the default dir so composer 
+    # doesn't error out.   
+    - name: Set writable perms on the default/ directory 
+      file: 
+        path: "{{ drupal_composer_install_dir }}/web/sites/default/"
+        mode: u+rw
+        recurse: yes
+        owner: "{{ ansible_user }}"
+      when:  defaultdir_exists.stat.exists == true
+      become: yes
+


ansible.cfg
```diff
@@ -12,5 +12,8 @@ internal_poll_interval = 0.001
 command_warnings = False
 deprecation_warnings = False
 
+# uncomment to speed up the install a bit by squashing package installations
+#squash_actions = apk,apt,dnf,homebrew,pacman,package,pkgng,yum,zypper
```
whikloj: This is new and exciting and I would like to subscribe to your newsletter. 

Under what circumstances would we _not_ want to squash package installations?

jhujasonw: @whikloj it's a "controversial" option.  There's talk of it even being deprecated in ansible 2.11.. I'm saddened by that, but certain packagers can't handle it right.  So if you use the `package` module to install packages instead of the OS specific manager modules, it *may* not work right.  I figured that disabling it by default would be a better option than fielding the questions about why things were failing if it were deprecated or what have you.  Kind of a "if you know about it, here it is" option, but a little more forward by including it in the ansible.cfg in the repo.

jhujasonw: Another thought is that it is disabled by default in the stock ansible.cfg, that was another reason for not enabling it by default.

whikloj: Ohhh ok, that makes sense.

jhujasonw:
Just wanted to see if there was anything else that was needed for this PR.  Please let me know if any adjustments are needed or if there's anything else that you would need from me.

dannylamb:
I tested this with the sandboxes for iCamp Arizona and it worked as advertised, so let's pull this in. :+1:

@manez Do we have a corporate CLA for JHU that covers @jhujasonw?

jhujasonw:
@dannylamb You don't have a corporate one but you have an individual one. I'm listed on your CLA page. https://github.com/Islandora/islandora/wiki/Contributor-License-Agreements

PR #124, opened 2019-07-11
bring role configs to inventory

Natkeeran:
!!! Please hold on testing this until the dependencies get merged. 

This PR depends on the following getting merged:
* https://github.com/Islandora-Devops/ansible-role-alpaca/pull/7
* https://github.com/Islandora-Devops/ansible-role-crayfish/pull/28

# What does this Pull Request do?
This PR make some of the variables that need to be configured for multi sever deployment explicit in the inventory.

# How should this be tested?
* After the dependent PRs are merged, get this PR
* Vagrant up
* Sanity checks

# Interested parties
 @Islandora-Devops/committers

PR #167, opened 2020-02-14
Simplified Docker dev setup

birkland:
Draft PR to demonstrate Islandora 8 in Docker, using a simplified setup containing an external Ansible variables file, and _no_ special Docker inventory, in contrast to #124

Creates and runs two Docker containers:

* An `ansible` container that, upon startup, runs ansible to install Islandora 8 to the `vm` container
* A `vm` container that plays the role of an empty Centos 7 box upon which Islandora 8 is installed via the `ansible` container.

## Motivation
* Providing the start of a development environment that would allow modeling of "services deployed on separate hosts" as containers.  
  * Right now, though, everything is installed to a single container whose role is analogous to the Vagrant VM.  It's a start, though.
* Because using Ansible + Vagrant + Virtualbox in Windows is possible, but impractically awkward

## To Test
* Make sure Docker is installed, as well as `docker-compose` (usually comes with the Docker distribution, e.g on Mac and Windows)
* Make sure the Islandora ports are free on your machine (e.g. there isn't another Islandora running on the same ports using Vagrant)
* Check out this PR, and in the repository's directory (`claw-playbook/`) do `docker-compose build`
* Do `docker-compose up -d`
* Watch the logs of the `ansible` container via ` docker logs -f ansible`.  That's the ansible output!  
* At this point, the only way to know it's done/ready is to look at the logs.  So watch the logs until Ansible is finished
  * Once the Ansible stops, the container will just run infinitely tailing /dev/null.  The container won't exit if it succeeds or fails.  It runs infinitely because that has been helpful for me in debugging Ansible issues.
* If Ansible succeeds, try logging into Islandora at the usual address `localhost:8000`, and poke around!  

## Tips and tricks

* You can hop onto any container while it's running by _docker exec -it <name> /bin/bash_.  For example, `docker exec -it vm /bin/bash`
* `docker-compose stop` will stop all containers, but keep state.  You can later on do a `docker-compose up vm` just to start the Islandora "vm" container. 
  * I haven't yet put the persisted state into volumes, so files/state are just written in the container.  As a result, all state disappears in a _poof_ once the containers are destroyed, e.g. via `docker-compose down`.  `docker-compose stop` preserves state.

### DIFF
--- a/docker/vm/sestatus
+++ b/docker/vm/sestatus
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+echo "Current mode:                   permissive"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,41 @@
+version: '3.1'
+
+services:
+
+  vm:
+    build:
+      context: ./docker/vm
+    container_name: vm
+    ports:
+      - 8000:8000
+      - 8080:8080
+      - 3306:3306
+      - 5432:5432
+      - 8983:8983
+      - 8161:8161
+    networks:
+      - islandora
+    privileged: true  
+    volumes:
+      - keys:/root/.ssh
+
+  ansible:
+    container_name: ansible
+    build:
+      context: ./docker/ansible
+    networks:
+      - islandora
+    depends_on: 
+      - vm
+    volumes:
+      - keys:/root/.ssh
+      - ./:/root/playbook:Z
+  
+networks:
+  islandora:
+    driver: bridge
+
+volumes:
+  keys:
+    driver: local
+    
--- a/docker/ansible/Dockerfile
+++ b/docker/ansible/Dockerfile
@@ -1,0 +1,17 @@
+FROM centos:7.6.1810
+
+ENV ISLANDORA_DISTRO="centos/7" \
+    ANSIBLE_CONFIG=/root/playbook/ansible.cfg 
+
+WORKDIR /root/playbook
+
+RUN yum -y install epel-release && \
+    yum -y install git openssl openssh-clients python-pip python-devel python
+
+RUN pip install ansible==2.8.7
+
+COPY entrypoint.sh /bin/
+
+RUN chmod 700 /bin/entrypoint.sh
+
+ENTRYPOINT [ "/bin/entrypoint.sh" ]
--- a/docker/ansible/entrypoint.sh
+++ b/docker/ansible/entrypoint.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+ssh-keyscan -H vm > /root/.ssh/known_hosts
+
+ansible-galaxy install --role-file=/root/playbook/requirements.yml --roles-path=roles/external
+
+ansible-playbook \
+    -u root \
+    -i ./inventory/vagrant/hosts \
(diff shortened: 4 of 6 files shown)


whikloj:
Trying this out as the normal method is failing and I have a problem with the ansible box.
```
Step 6/8 : COPY entrypoint.sh /bin/
 ---> ce3d6680bac4
Step 7/8 : RUN chmod 700 /bin/entrypoint.sh
 ---> Running in 10142b8583cb
Removing intermediate container 10142b8583cb
 ---> c27c6cded545
Step 8/8 : ENTRYPOINT [ "/bin/entrypoint.sh" ]
 ---> Running in 3bb0f1944e01
Removing intermediate container 3bb0f1944e01
 ---> fa0a1fe6d348
Successfully built fa0a1fe6d348
Successfully tagged clawplaybook_ansible:latest
WARNING: Image for service ansible was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Creating vm ... 
Creating vm ... done
Creating ansible ... 
Creating ansible ... error

ERROR: for ansible  Cannot start service ansible: cgroups: cannot find cgroup mount destination: unknown

ERROR: for ansible  Cannot start service ansible: cgroups: cannot find cgroup mount destination: unknown
ERROR: Encountered errors while bringing up the project.
```

whikloj:
Ok I forgot to run `docker-compose build` first, so I stopped and removed all running containers. 
Then I removed the images.
Then ran `docker-compose build`
```
> docker-compose build
Building vm
Step 1/10 : FROM centos:7.6.1810
 ---> f1cb7c7d58b7
Step 2/10 : ENV container docker
 ---> Running in 049a03d4df43
Removing intermediate container 049a03d4df43
 ---> 61e287a5b98d
Step 3/10 : COPY sestatus /bin/
 ---> 64d85322a382
Step 4/10 : RUN chmod 700 /bin/sestatus
 ---> Running in 38df5cc95229
ERROR: Service 'vm' failed to build: cgroups: cannot find cgroup mount destination: unknown
```

docker/vm/sestatus
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+echo "Current mode:                   permissive"
```
whikloj: Does this actually disable SELinux or just the reporting functions?

birkland: Since SELinux isn't really running in Docker containers, this just tricks Ansible into passing its "is selinux set up correctly?" test.

whikloj:
@birkland So I have been able to get this to work, but even though the tomcat8 role shows no errors it seems to fail starting up. 
```
Offending resource: ServletContext resource [/WEB-INF/classes/spring/repository.xml]; nested exception is org.springframework.beans.factory.BeanDefinitionStoreException: IOException parsing XML document from URL [file:/opt/fcrepo/configs/fcrepo-config.xml]; nested exception is java.io.FileNotFoundException: /opt/fcrepo/configs/fcrepo-config.xml (No such file or directory)
 ```
I see the files now, and I can restart the service. It seems like at the time the files are not there and the service never restarts. Ideas how to debug this?

PR #174, opened 2020-04-08
Use roles/external as the default location for ansible-galaxy install

nigelgbanks:

If a user executes the following command:

```
ansible-galaxy install -r requirements.xml
```

The roles will be installed into **roles/internal**, since no specific path has been specified and the first path listed in the configuration is **roles/internal**. 

This pull request changes this behaviour such that the default location it will install roles into is **roles/external**. This is useful for those who want to use the Ansible roles but do not with to use it via Vagrant (which specified the **roles/external** path explicitly).

# What's new?
A in-depth description of the changes made by this PR. Technical details and
 possible side effects.

* Default location of roles installation changed to **roles/external** rather than **roles/internal**

# How should this be tested?

Execute the command:

```
ansible-galaxy install -r requirements.xml
```

Note that the roles are installed in the correct location without the need of also giving the explicit path.

# Interested parties
@Islandora-Devops/committers

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,7 +52,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     ansible.compatibility_mode = "auto"
     ansible.playbook = "playbook.yml"
     ansible.galaxy_role_file = "requirements.yml"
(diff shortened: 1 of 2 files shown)


dannylamb:
Can't tell you how many times this has annoyed me.

PR #182, opened 2020-07-02
Allow disabled selinux

whikloj:
**GitHub Issue**: https://github.com/Islandora/documentation/issues/1556

Accounts for SELinux having a status of disabled as well as permissive and enforcing.

# How should this be tested?

Really this needs a box which starts with a disabled SELinux and then run the playbook over-top of it. In which case it should not make any changes.

# Interested parties
@Islandora-Devops/committers and @ainsofs


dannylamb:
@ainsofs Any chance you can try this out and let us know if it works for you?

PR #185, opened 2020-08-28
Master -> Main Rename

dannylamb:
Part of https://github.com/Islandora/documentation/issues/1595

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ You can access the IIIF interface at: http://localhost:8080/cantaloupe/iiif/2/
 ### JWT
 
 Islandora 8 uses JWT for authentication across the stack. Crayfish microservices, Fedora, and Drupal all use them. 
-Crayfish and Fedora have been set up to use a master token of `islandora` to make testing easier. To use it, just set
+Crayfish and Fedora have been set up to use a main token of `islandora` to make testing easier. To use it, just set
 the following header in HTTP requests:
 
   * `Authorization: Bearer islandora`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -98,7 +98,7 @@ You can access the IIIF interface at: http://localhost:8080/cantaloupe/iiif/2/
 ### JWT
 
 Islandora 8 uses JWT for authentication across the stack. Crayfish microservices, Fedora, and Drupal all use them. 
-Crayfish and Fedora have been set up to use a master token of `islandora` to make testing easier. To use it, just set
+Crayfish and Fedora have been set up to use a main token of `islandora` to make testing easier. To use it, just set
```
rosiel, line 101: were these tokens changed? I'm guessing not, as this is a deep-within-the-stack use and not a github-branch-related one?

whikloj: I think the only thing to change is the term "master" to "main" in the above sentence. The token itself is still "islandora" and behaves the same way.

rosiel: in this context, 'main' doesn't seem to convey the same meaning. Would 'universal' or 'global' or 'generic' be applicable?

manez: Does "default" make sense in this case?

whikloj: I'm good with either.

rosiel: Default sounds good.

dannylamb: yeah wasn't sure about that when doing this.  i'm good with default :+1:

dannylamb:
Should be good to go now @manez @rosiel @whikloj

rosiel (approved):

PR #198, opened 2021-03-31
Ubuntu 20.04 Focal Fossa.

alxp:
**GitHub Issue**: [Update playbook to Ubuntu 20.04 Focal64](https://github.com/Islandora/documentation/issues/1792)

Base Ubuntu version on 2020-04 FocalFossa, includes an update to Tomcat 9 and other roles.

# What's new?

Updated several external Islandora-Devops roles to be compatible with Ubuntu 20.04.

No

The Islandora-Devops.tomcat8 role may be better renamed to tomcat9 or we go down the route to make it be either.

This makes existing installations based on the playbook to be upgraded to a newer OS version. In place upgrades will need to be tested. 

# How should this be tested?

Run a full playbook installation, and tests to make sure object creation work as expected.

### Testing updated Ansible roels

This PR has been updated so that requirements.yml points to where the updated roles will live in the Islandora-Devops GitHub project. To test them before they are merged, roll back this PR's branch by 1 commit to have the requirements point to updated branches in https://github.com/roblib.

Test updating an existing playbook installation.

# Additional Notes:

### Dropping CentOS

> Do we still use CentOS? I noticed that building based on Centos does not work for Drupal 9 as the MySQL version is too old. Is this used by anyone going forward?

It was decided in [March 31, 2021 meeting](https://github.com/Islandora/documentation/wiki/March-31%2C-2021) that CentOS should no longer be supported due to lack of use and RedHat announcing that they were moving away from using it. References to CentOS were removed from the README in this PR but it hasn't been completely removed from the playbook and roles yet.

### Backwards compatibility question:

> The last update went from Xenial to Bionic without attempting to preserve the ability to build on Xenial. Is this ok to do again or should we add variables to allow for backwards compatibility?

It was decided in the [March 31, 2021 Meeting](https://github.com/Islandora/documentation/wiki/March-31%2C-2021) call that we should continue to move forward and support latest versions as the Playbook was never properly set up as an Ansible deployment platform.

# Interested parties

@Islandora-Devops/committers

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,8 +11,8 @@ $memory = ENV.fetch("ISLANDORA_VAGRANT_MEMORY", "4096")
 $hostname = ENV.fetch("ISLANDORA_VAGRANT_HOSTNAME", "islandora8")
 $virtualBoxDescription = ENV.fetch("ISLANDORA_VAGRANT_VIRTUALBOXDESCRIPTION", "Islandora 8")
 
-# Available boxes are 'islandora/8', ubuntu/bionic64' and 'centos/7'
-# Use 'ubuntu/bionic64' or 'centos/7' to build a dev environment from scratch.
+# Available boxes are 'islandora/8' and 'ubuntu/focal64'
+# Use 'ubuntu/focal64' to build a dev environment from scratch.
 # Use 'islandora/8' if you just want to download a ready to run VM.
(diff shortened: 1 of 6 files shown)


seth-shaw-unlv (approved):
I smoke tested an image and a document. All the pieces appear to be in working order. 👍

Just let me know when you're done cleaning things up.

seth-shaw-unlv (approved):
👍

PR #210, opened 2022-03-30
Fixes for running provision on an existing site.

alxp:
**GitHub Issue**: (link)

[#2070: Updating a site via the Playbook fails due to field_abstract type mismatch](https://github.com/Islandora/documentation/issues/2070)

Also fixes error where Grok clone fails with error "Would clobber existing tag."

* Other Relevant Links (Google Groups discussion, related pull requests,
 Release pull requests, etc.)

# What does this Pull Request do?

Fixes so that the playbook will run without error on an existing site. 
Fixes a number of errors encountered when running the playbook on an existing Islandora installation.

# What's new?

 - Do not attempt to re-import features
 - Fix grok clone error "This would clobber an existing branch".

* Does this change require documentation to be updated? 
* 
  no (No where is the behaviour of re-importing the features documented)

* Does this change add any new dependencies?  
  no

* Does this change require any other modifications to be made to the repository
  no

* Could this change impact execution of existing code?

No

# How should this be tested?

If available, run the playbook on an existing Islandora site with the standard install profile.

Also run the playbook as a fresh installation.

Make sure derivative generation works as expected.

# Additional Notes:

Note that this PR includes the recent change to move from Karaf to the one-jar Alpaca, so this testing may uncover bugs relating to that update.

# Interested parties
Tag (@ mention) interested parties or, if unsure, @Islandora-Devops/committers

PR #211, opened 2022-04-07
Target ids directly as opposed to the group.

jordandukart:
**GitHub Issue**: N/A

Migrate_tools was removed in https://github.com/Islandora/islandora/pull/859 to use Drush's built in runners. This targets the IDs directly as opposed to the non-existent `--group` parameter.


alxp:
Tested with PR https://github.com/Islandora-Devops/islandora-playbook/pull/210 and got past the migrate error and successfully ran the playbook as an update.

alxp (approved):

PR #216, opened 2022-05-04
Cantaloupe Support JP2 out of the box

alxp:

The Cantaloupe image viewer was not properly set up to support JP2 out of the box. 

# What's new?

Added a few configurations and packages to load OpenJP2.
OPenJP2 tools Ubuntu package
Explicitly set manual selection strategy for Cantaloupe 
Use OpenJP2 rather than Grok. Keeping Grok as it is used to generate derivatives

OpenJP2-tools Ubuntu packages

no

No

# How should this be tested?

From fresh playbook install:

1. Create a Repository Item object of type Page
2. Add Media of type File with a JP2, and tagged as Original File.
3. Copy the URL where the JP2 gets put, should be a /_flysystem URL 
4. URL-encode the image URL, e.g.,
```
$ php
<?php 
print(urlencode($url)):
```
5. GO to the Cantaloupe IIIF endpoints with this encoded URL:

http://localhost:8080/cantaloupe/iiif/2/{identifier}/info.json
http://localhost:8080/cantaloupe/iiif/2/{identifier}/full/full/0/default.jpg

You should see the JSON info and a tiled version of the image at those URLs.

# Interested parties

### DIFF
--- a/inventory/vagrant/group_vars/tomcat.yml
+++ b/inventory/vagrant/group_vars/tomcat.yml
@@ -69,18 +69,18 @@ cantaloupe_deploy_war_path: "{{ tomcat9_home }}/webapps"
 cantaloupe_user: tomcat
 cantaloupe_group: tomcat
 cantaloupe_admin_enabled: "true"
-cantaloupe_GrokProcessor_path_to_binaries: /user/local/bin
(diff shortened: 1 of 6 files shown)


jordandukart (approved):
Works as intended.

PR #217, opened 2022-06-08
Allow-list composer plugins

alxp:
**GitHub Issue**: (link)

 Release pull requests, etc.)

Discussion of composer's plans to disallow plugins by default starting in July 2022: 

https://www.drupal.org/project/drupal/issues/3255749

Explicitly allows composer plugins that Drupal uses.

# What's new?

grab latest releases.

Add the composer plugins Drupal uses to an explicit allow list to avoid possible breaking behaviour starting July 2022.

* Does this change require documentation to be updated? 

No

No

No

* 
Existing Drupal sites should not be affected. New installs will default to stable version of Drupal core.

# How should this be tested?

Run a fresh playbook install.

Test object creation to ensure derivatives are generated.

Run a composer command while logged in as vagrant. You should not get a warning about allowed plugins

# Additional Notes:

The 'acl' Ubuntu package was added to get around a problem with newer versions of Ansible when running as a less-privileged user. See:  https://github.com/ansible/ansible/issues/74830

# Interested parties

### DIFF
--- a/inventory/vagrant/group_vars/all/main.yml
+++ b/inventory/vagrant/group_vars/all/main.yml
@@ -14,6 +14,7 @@ islandora_extra_ubuntu_packages:
   - build-essential
   - vim
   - python3-mysqldb
+  - acl
 
 islandora_extra_centos_packages:
   - wget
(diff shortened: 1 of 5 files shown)


rosiel:
The composer-plugins-fix task never seemed to work. We're reducing the need for this with #221 .

PR #218, opened 2022-06-08
Composer stability fix to get Drupal 9.3.x.

alxp:

Set composer default options to opt for stable versions by default.

# What's new?

The playbook was using geerlingguy.composer's default composer tags which included prefer-dist, this was causing composer create-project to get the development version of drupal/recommended-project.

We just needed to explicitly specify Flysystem  as a beta since it has no stable release.

* Does this change require documentation to be updated?  no 
* Does this change require any other modifications to be made to the repository no
No, existing drupal installs aren't replaced by the playbook.

# How should this be tested?

Run a fresh playbook install.

Verify that the version of Drupal core is 9.3.x (as of June 8 2022).

Verify derivative creation.

# Interested parties

### DIFF
--- a/inventory/vagrant/group_vars/webserver/drupal.yml
+++ b/inventory/vagrant/group_vars/webserver/drupal.yml
@@ -1,6 +1,7 @@
 ---
 
 drupal_composer_install_dir: /var/www/html/drupal
+drupal_composer_project_options: "--prefer-dist --no-interaction"
 drupal_core_owner: "{{ ansible_user }}"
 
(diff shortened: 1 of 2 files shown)


whikloj:
Standing this up today (June 24, 2022) I ended up with 9.4.1 is that correct or how can I verify?

alxp:
Drupal 9.4 was released this week so yes that's the expected behaviour.

whikloj (approved):
Works as described 👍

PR #219, opened 2022-07-04
Remove hard-coded Solr version when generating Drupal solr config.

alxp:

Make the Drupal solr config version match the installed version of Solr.

# What's new?

When generating the solr_config.zip file, the Drush command now uses the {{ solr_version }} from the inventory rather than the hard-coded "7.1".

* Does this change require documentation to be updated?  No
* Does this change add any new dependencies?  No
 (i.e. Regeneration activity, etc.)?  No
* Could this change impact execution of existing code? No

# How should this be tested?

Run the playbook.

Go to  /admin/config/search/search-api/server/

Click on the Default Solr Server and default index links. Check that Drupal does not print a warning about  out-of-date solr config files 

# Interested parties

### DIFF
--- a/inventory/vagrant/group_vars/solr.yml
+++ b/inventory/vagrant/group_vars/solr.yml
@@ -1,4 +1,4 @@
-solr_version: "8.11.1"
+solr_version: "8.11.2"
 
 solr_cores:
   - ISLANDORA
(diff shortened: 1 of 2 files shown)


rosiel (approved):
Tested and works. Installs 8.11.1.

PR #126, opened 2019-07-29
separate solr tasks so it can be deployed to its own server

Natkeeran:
search_api_solr 8.3 is not quite working, thus this PR is stuck: https://github.com/Islandora-Devops/claw-playbook/pull/123.  Want to get the ansible changes to support indepndent solr server deployement in, thus this PR.  Will circle back to the search_api_solr 8.3 issue later on.

**GitHub Issue**: (link)
https://github.com/Islandora-CLAW/CLAW/issues/1197

# What does this Pull Request do?
* Supports installing solr on its own server.

# How should this be tested?
* Get the PR
* vagrant up
* Create content
* Verify that it gets indexed and returned 

In addition, testing the following is helpful.
* Get the PR
* Modify or create the following playbook:
```
---

- include: bootstrap.yml
  tags:
    - bootstrap
- include: solr.yml
```
* deploy and check that solr has been correctly installed

# Interested parties
@Islandora-CLAW/committers

PR #226, opened 2022-09-09
Starter site integration

adam-vessey:
**GitHub Issue**:
* https://github.com/Islandora/islandora-starter-site/issues/2
* https://github.com/Islandora/islandora_defaults/issues/74

 Release pull requests, etc.)
    * [Code Chill: Islandora Defaults](https://groups.google.com/g/islandora/c/uGzhTnW4TUI)
    * [Islandora Foundation contracting with DGI for "Modularizing Islandora" work](https://groups.google.com/g/islandora/c/vO1JAPO7GXM)

Gets the new "starter site" project installable via ansible.

# What's new?

Added the `starter` and `starter_dev` profiles, with minor reworks to allow some tasks to be suppressed.

* Does this change require documentation to be updated? Unknown.
* Does this change add any new dependencies? No/only if using the new "starter" profile(s).
 (i.e. Regeneration activity, etc.)? No.
* Could this change impact execution of existing code? Potentially, with how some of the ansible state had to be manipulated... it seems some things were referring to the wrong state (such as the Matomo things defining the `set_search_api_config` value, and "Solr" stuff still referring to it). Also, if there's other 

# How should this be tested?

Smoketest and compare the behaviours of the `standard` and `demo` "profiles" from before and after the changes made here, they should behave the same as they did previously.

Test out the new `starter` and `starter_dev` "profiles". Should come up and look much the same, with the only difference presently being that `_dev` performs a clone of the project (which is only really evident when looking at how it the project exists as it was deployed on the filesystem).

# Additional Notes:

Nothing to mind...

# Interested parties


roles/internal/webserver-app/tasks/drupal.yml
```diff
@@ -21,11 +21,53 @@
...
+
+- name: Content Sync Directory configuration
+  blockinfile:
+    state: present
+    block: |
+      global $content_directories;
+      $content_directories['sync'] = $app_root.'/../content/sync';
+    path: "{{ drupal_site_default_settings }}"
+    marker: // {mark} ANSIBLE MANAGED BLOCK - content sync dir
+  when: webserver_app_do_content_sync_config
```
adam-vessey, line 63: Not sure what the best approach was here, to allow for more granular control over what was being done in what was originally called the "Trusted Host Settings" bit, which was doing more than just setting the "trusted hosts" settings... elected to keep the original, but additionally have all the bits it contained be available in separate chunks, as we don't really want the `content/sync` thing being done in the `starter` "profile".

roles/internal/webserver-app/tasks/drupal.yml
```diff
@@ -48,24 +90,28 @@
...
+  when: webserver_app_configure_matomo
 
-- name: Set Matmo sever url.
+- name: Set Matomo server url.
   command: "{{ drush_path }} --root {{ drupal_core_path }} -y config-set matomo.settings url_http http://localhost:8000/matomo/"
-  register: set_search_api_config
-  changed_when: "'Do you want to update site_id' in set_search_api_config.stdout"
+  register: set_matomo_server_url_config
+  changed_when: "'Do you want to update site_id' in set_matomo_server_url_config.stdout"
+  when: webserver_app_configure_matomo
```
adam-vessey: The `register`/`changed_when` things touching `set_search_api_config` seems like something of copypasta error, as we're not dealing with "search API" here... so adjusted things in a way that seems more consistent?

... it's... _possible_ that the second will have to have the string it's searching for changed... instead of `site_id`, to be `url_http` or something of the like.

vars/starter.yml
```diff
@@ -0,0 +1,24 @@
+---
+drupal_build_composer_project: true
+drupal_composer_dependencies: []
+# Presently pointing at the main branch.
+drupal_composer_project_package: 'islandora/islandora-starter-site:dev-main'
```
adam-vessey, line 8: This should eventually be pointed at whatever reference structure we move to use for releases... or maybe left without a version reference to pull whatever "latest" version?

vars/starter_dev.yml
```diff
@@ -0,0 +1,34 @@
+---
+drupal_build_composer_project: false
+drupal_deploy: true
+# XXX: Just pointing at my repo for the moment, for testing... should be pointed
+# back at Islandora's prior to merging.
+drupal_deploy_repo: https://github.com/adam-vessey/islandora-starter-site.git
+# XXX: Should be made to point at "main".
+drupal_deploy_version: implementation
```
adam-vessey: Yeah, needs to be pointed at the merged bits, whenever it's possible to do so.

vars/starter.yml
```diff
@@ -0,0 +1,24 @@
+---
+drupal_build_composer_project: true
+drupal_composer_dependencies: []
```
adam-vessey: Strictly, doesn't need to be specified here (the default in the `geerlingguy.drupal` is an empty list); however, thought to leave it to maintain something of a parallel structure with the `demo` and `standard` "profiles"... could be useful for people looking to hack on things?

vars/starter.yml
```diff
@@ -0,0 +1,24 @@
...
+# Presently pointing at the main branch.
+drupal_composer_project_package: 'islandora/islandora-starter-site:dev-main'
+
+# XXX: Strictly, irrelevant due to using the `--existing-config` flag.
+drupal_install_profile: minimal
+
+# None necessary, due to using an existing configuration.
+drupal_enable_module: []
```
adam-vessey: Same deal: Strictly, doesn't need to be specified here (the default in the `geerlingguy.drupal` is an empty list); however, thought to leave it to maintain something of a parallel structure with the `demo` and `standard` "profiles"... could be useful for people looking to hack on things?

vars/starter.yml
adam-vessey: Adjusted, to leave with a comment and with it commented out.

vars/starter.yml
adam-vessey: Adjusted, to leave with a comment and with it commented out.

roles/internal/webserver-app/tasks/solr.yml
```diff
@@ -6,12 +6,12 @@
...
   command: "{{ drush_path }} --root {{ drupal_core_path }} -y config-set search_api.server.default_solr_server backend_config.connector_config.host {{ hostvars[groups['solr'][0]].ansible_host  }}"
   register: set_search_api_config_host
-  changed_when: "'Do you want to update' in set_search_api_config.stdout"
+  changed_when: "'Do you want to update' in set_search_api_config_host.stdout"
 
 - name: Set default solr server to point to first core
   command: "{{ drush_path }} --root {{ drupal_core_path }} -y config-set search_api.server.default_solr_server backend_config.connector_config.core {{ solr_cores[0] }}"
   register: set_search_api_config_core
-  changed_when: "'Do you want to update' in set_search_api_config.stdout"
+  changed_when: "'Do you want to update' in set_search_api_config_core.stdout"
```
adam-vessey, line 14: Looking back at https://github.com/Islandora-Devops/islandora-playbook/pull/126/files#diff-5bddb401e18178b7584b51cd9831eaa348385079b2224a85e3db8fa2412fc547L5-R15 ... I believe this is the intent, as otherwise, it seems that the `set_search_api_config` bit would've been being registered by the "Matomo" setup, in the `webserver-app/tasks/drupal.yml` bit?

vars/starter_dev.yml
adam-vessey: Done in https://github.com/Islandora-Devops/islandora-playbook/pull/226/commits/a18bde06eaa5a4b0e54c3b87b6e400a716184c1d

roles/internal/webserver-app/tasks/solr.yml
rosiel: Yeah that seems like a typo in the Matomo stuff.

vars/starter.yml
```diff
@@ -0,0 +1,27 @@
...
+# XXX: Strictly, irrelevant due to using the `--existing-config` flag.
+drupal_install_profile: minimal
+
+# None necessary, due to using an existing configuration.
+#drupal_enable_modules: []
+
+drupal_site_install_extra_args:
+  - '--existing-config'
+
+openseadragon_composer_require: false
```
rosiel, line 21: Some of these - setting up Matomo and FITS - seem like they're `false` only because theyre out of scope at the moment but we probably eventually want them. Can you denote/comment/group them in some way to distinguish from the others which we do/don't need due to the architecture of the Starter Site?

adam-vessey: Took a shot in https://github.com/Islandora-Devops/islandora-playbook/pull/226/commits/52f962aa6aad7901370e2d634c40991c51efe198 ... thoughts?

adam-vessey: I feel like I should mention: Anything that _does_ need to do any configuration probably maybe shouldn't do it by `config:set`'ing anything, but instead to use other mechanisms such as `config_override` to avoid potentially dirtying configs... to avoid specific environmental concerns creeping into things; for example: If the `matomo.settings:url_http` value was made to dynamically point at the hostname instead of just specifying `localhost:8000`, then yeah, it wouldn't really belong in the config that gets used between environments where that remote would(/could) vary.

... This... _does_ kind of get into a weird area with [how the JWT stuff is presently configured](https://github.com/Islandora-Devops/islandora-playbook/blob/311aa2126947869ce20d7765e9689de3d219d835/roles/internal/webserver-app/tasks/jwt.yml#L10-L39), in that because things were based on the "standard" playbook install, that it just so happens to match the paths; however, if other paths are ever specified for `webserver_app_jwt_key_path`, then it could end up diverging...

roles/internal/webserver-app/tasks/drupal.yml
adam-vessey: > ... it's... _possible_ that the second will have to have the string it's searching for changed... instead of `site_id`, to be `url_http` or something of the like.

Did indeed appear to be the case. Running the command manually: 
Screen Shot 2022-09-13 at 12 10 52 PM

Addressed in https://github.com/Islandora-Devops/islandora-playbook/pull/226/commits/dd550679f02cef9a3bc1961f2efc97d1981033fe

roles/internal/webserver-app/tasks/drupal.yml
```diff
@@ -1,6 +1,6 @@
 ---
 
-- name: Trusted Host Settings
+- name: Trusted Host Settings and more
```
bibliophileaxe, line 3: Maybe we can rename this to settings.php settings or something generic?

adam-vessey: There's more than just this singular task doing `settings.php` things... really, was just trying to reflect the reality in the name that it was doing more than just the "trusted host" stuff it was originally named... Like the flysystem and `$content_directories` stuff... and trying to follow something of the established patterns of naming things.

vars/starter_dev.yml
```diff
@@ -0,0 +1,32 @@
+---
+drupal_build_composer_project: false
+drupal_deploy: true
+#drupal_deploy_repo: https://github.com/Islandora/islandora-starter-site.git
+#drupal_deploy_version: main
+drupal_deploy_repo: https://github.com/adam-vessey/islandora-starter-site.git
+drupal_deploy_version: feature/playbook-standard-replacement
```
adam-vessey: Back to needing to be pointed back upstream after https://github.com/Islandora/islandora-starter-site/pull/6 is merged.

adam-vessey: Addressed in https://github.com/Islandora-Devops/islandora-playbook/pull/226/commits/b3f1bca7b8f8dfe18f16b10e60e259e3a984395e

willtp87 (approved):

rosiel (approved):

PR #239, opened 2022-12-06
Set crayfits url using variables for remote deployment.

rosiel:
**why**: 

Alpaca gets configured to look for fits at `http://localhost:8000/crayfits`. This adds the URL to the inventory variables, so it gets set to the appropriate value.

Deploy to a non-localhost machine. 
Before this: no FITS derivatives and lots of errors in `/opt/alpaca/logs/alpaca.log`. 
After: hopefully you get FITS derivatives.


alxp (approved):
I had run into this earlier, but never got around to pushing the fix up. Thanks for this.

PR #240, opened 2022-12-09
Add tempDirectory to Homarus config.

rosiel:
**GitHub Issue**: none

A recent commit of Crayfish added a new parameter to the Homarus services.yaml config file.
When deploying with Vagrant, the default config file gets overwritten with one "managed by ansible" and is generated from a template.

Because of the missing variable, i was getting this error: ` Uncaught Exception: Cannot autowire service "App\\Islandora\\Homarus\\Controller\\HomarusController": argument "$tempDirectory" of method "__construct()" is type-hinted "string", you should configure its value explicitly.` and no homarus derivatives were being created.

This pull updates the template so it includes the new `$tempDirectory` variable and makes it configurable. Though it does not set a default value in the Vagrant inventory.

Makes homarus run! 

# What's new?

`tempDirectory` parameter in the Homarus settings.yaml file.

# How should this be tested?

Replicate:
* spin up playbook (use Crayfish on 3.x, which is currently default for Vagrant)
* Try to make an audio or video derivative that uses Homarus. No dice.

With this fix:
* Spin up playbook
* Try to make an audio or video derivative that uses Homarus. Success!

# Additional Notes:
 PR.

# Interested parties
@alxp @whikloj

### DIFF
--- a/inventory/vagrant/group_vars/crayfish.yml
+++ b/inventory/vagrant/group_vars/crayfish.yml
@@ -1,5 +1,4 @@
 ---
-crayfish_version_tag: 3.x
 crayfish_db: "{{ islandora_db }}"
 
 crayfish_fedora_base_url: "http://{{ hostvars[groups['tomcat'][0]].ansible_host }}:8080/fcrepo/rest"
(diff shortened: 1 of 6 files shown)


whikloj:
I haven't tested this yet, but it makes sense. 

I do wonder if we are now keeping most of these roles internal, can we just agree to update the versions in one spot.

Either leave the original version in the role and update the inventory/vagrant override or remove the inventory/vagrant override and just update the role version.

rosiel:
Good point, I'll just do it in the role. I think that's much cleaner for the future.

alxp (approved):

alxp:
Tested this and it works as advertised.

PR #243, opened 2022-12-16
Upgrade to jammy64 and PHP 8.1

whikloj:
**GitHub Issue**: n/a

Upgrades the normal base box to `ubuntu/jammy64` and the version of PHP to 8.1

# What's new?

Centralize the `php_version` variable in one place in `inventory/vagrant/group_vars/all/main.yml` and change the base box.

Also deletes the database tests from Crayfish as Crayfish no longer uses a database.

# How should this be tested?

**Note**: I have updated some of the ansible roles, so empty your `roles/external` directory before starting this process.

You need to remove the existing `islandora_base` and rebuild the base box, then destroy the base box image and build a new box using the base box.

```
> rm islandora_base

> ISLANDORA_BUILD_BASE=true vagrant up

> vagrant package --output=islandora_base

> ISLANDORA_BUILD_BASE=true vagrant destroy

> vagrant up
```

# Interested parties
Tag (@ mention) interested parties or, if unsure, @Islandora-Devops/committers especially @rosiel


webserver.yml
```diff
@@ -15,15 +12,15 @@
...
-      when: islandora_build_base_box|bool == True
+      when: islandora_build_base_box|bool == False
     - name: geerlingguy.php-mysql
-      when: islandora_build_base_box|bool == True
+      when: islandora_build_base_box|bool == False
     - name: geerlingguy.git
       when: islandora_build_base_box|bool == True
     - name: geerlingguy.composer
-      when: islandora_build_base_box|bool == True
+      when: islandora_build_base_box|bool == False
```
rosiel: Did you intend to move PHP and Composer to the second step, or was this for easier testing? (either's ok, just checking)

rosiel: Ah! it is causing a problem. My playbook breaks while creating the base box. At line 18 of crayfish.yml we try to install composer, but we haven't installed php yet.

seth-shaw-asu:
~~I followed the instructions, but I still have PHP 7.4. I must have missed something.~~

_Disregard, it was user error._

webserver.yml
seth-shaw-asu: I ran into the same issue. I switched them back:
```diff
diff --git a/webserver.yml b/webserver.yml
index 171d9e9..28fb86c 100644
--- a/webserver.yml
+++ b/webserver.yml
@@ -12,11 +12,11 @@
     - name: geerlingguy.apache
       when: islandora_build_base_box|bool == True
     - name: geerlingguy.php-versions
-      when: islandora_build_base_box|bool == False
+      when: islandora_build_base_box|bool == True
     - name: geerlingguy.php
-      when: islandora_build_base_box|bool == False
+      when: islandora_build_base_box|bool == True
     - name: geerlingguy.php-mysql
-      when: islandora_build_base_box|bool == False
+      when: islandora_build_base_box|bool == True
     - name: geerlingguy.git
       when: islandora_build_base_box|bool == True
     - name: geerlingguy.composer
```
Now it appears to work as advertised.

seth-shaw-asu (changes requested):
Turning the PHP roles back to True makes it work as advertised.

whikloj:
There is a hiccup I discovered which is that once you use the base box once, it is stored in the vagrant box list under your home directory and subsequent builds of `ISLANDORA_BUILD_BASE=true vagrant up` will build a new box and you can package them, but that file does not update the cached `box` and consequently you still get the old base box when you try build a VM.

TL;DR 
If you are running `ISLANDORA_BUILD_BASE=true vagrant up` then you want to run `vagrant box remove islandora_base` before you run `vagrant up` to ensure it installs the newly built base box on disk.

seth-shaw-asu:
~~This wasn't an issue before, but now the geerlingguy.php is choking:~~
```
TASK [geerlingguy.php : Ensure PHP packages are installed.] ********************
Monday 19 December 2022  12:38:33 -0800 (0:00:01.157)       0:05:07.818 ******* 
fatal: [default]: FAILED! => {"changed": false, "msg": "No package matching 'php7.0-common' is available"}
```
~~As far as I can tell, geerlingguy only sets php7.0 as a default version for unversioned Debian, Debian-9, or Ubuntu-16. (See `roles/external/geerlingguy.php/vars/`.)~~

_I must have done a bad pull of the updates. I cleaned everything out and did a fresh application of the PR and now the base box builds. It's packaging right now but then I'll spin up the second stage and test the JP2._

seth-shaw-asu (approved):

Issue #248, opened 2023-01-12
Playbook is setting configs that it shouldn't and is skipping setting configs.


rosiel:
There's a whole [section of the post-install.yml file](https://github.com/Islandora-Devops/islandora-playbook/blob/dev/post-install.yml#L45-L60) that does `drush cset` commands (config set). If I understand correctly, these are a holdover from when we had Defaults, which potentially didn't include setting these config variables. This includes setting Media to have standalone URLs and whether we should remove ?_format=jsonld.

We now have a place for these Config decisions to go (starter site, and also install profile).

There are also a handful of variables that need to match the environment, such as the location of cantaloupe or fedora or whatnot. The values currently used in the Starter Site are values that will work with the playbook. However, it'd be good form for the Playbook to set these.

* remove these where they are already set in the starter site, and
* ensure that all the variables that need to be set by the playbook are being set by the playbook.

rosiel:
This has become a bit of a cleaning exercise regarding the use of Drush in the Playbook. 

### We can remove these from post-install.yml, because they're part of the starter site 
```
    - name: Set JSONLD Config
      command: "{{ drush_path }} --root {{ drupal_core_path }} -y cset --input-format=yaml jsonld.settings remove_jsonld_format {{ drupal_jsonld_remove_format }}"

    - name: Set media urls
      command: "{{ drush_path }} --root {{ drupal_core_path }} -y cset --input-format=yaml media.settings standalone_url true"
[...]
    - name: Set iiif manifest view
      command: "{{ drush_path }} --root {{ drupal_core_path }} -y cset --input-format=yaml openseadragon.settings manifest_view iiif_manifest"

```

### We do need to set these (but could rename to be clearer)

```
    - name: Set iiif url
      command: "{{ drush_path }} --root {{ drupal_core_path }} -y cset --input-format=yaml islandora_iiif.settings iiif_server {{ openseadragon_iiiv_server }}"

    - name: Force Alpaca host setting to override install profile.
      command: "{{ drush_path }} --root {{ drupal_core_path }} -y -l localhost:{{ apache_listen_port }} cset islandora.settings broker_url tcp://{{ hostvars[groups['alpaca'][0]].ansible_host }}:61613"

```
### Things we need to set but are not in post-install.yml:

#### Matomo

* in `roles/internal/webserver-app/tasks/drupal.yml`. 
* hardcodes localhost and site_id = 1

#### Solr
[...]

PR #261, opened 2023-03-03
Let site (starter or otherwise) set media urls.

rosiel:
**GitHub Issue**: in part, addresses #248 

 Release pull requests, etc.)

Doesn't set media urls to have "Standalone URLs". This configuration is part of the starter site / install profile / whatever site you're using. The playbook currently insists on overriding whatever value you have set there in your site, which I think it shouldn't do.

# What's new?

* Removed a line that sets the value to be true.
* Does this change require documentation to be updated?  Some screenshots might need updating if they show media pages.
 (i.e. Regeneration activity, etc.)?  no
* Could this change impact execution of existing code? no

# How should this be tested?

Without this PR:
* install the starter site using "starter_dev"
* once installed, check the config diff using `drush cex --diff`. You should see that the media URLs have been set to True. 

With this PR:
* the `drush cex --diff` should not include this, because the value wasn't set in a way that contradicts the starter site.
 
# Additional Notes:
 PR.

# Interested parties

### DIFF
--- a/post-install.yml
+++ b/post-install.yml
@@ -46,9 +46,6 @@
     - name: Set JSONLD Config
       command: "{{ drush_path }} --root {{ drupal_core_path }} -y cset --input-format=yaml jsonld.settings remove_jsonld_format {{ drupal_jsonld_remove_format }}"
 
-    - name: Set media urls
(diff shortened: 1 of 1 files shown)


rosiel:
Anyone have the time to check this one out?

seth-shaw-asu (approved):

PR #265, opened 2023-06-13
Fix version of ubuntu in README.

rosiel:
**GitHub Issue**: (link)

Since #243 we don't use focal, we use jammy.

# Interested parties


whikloj (approved):
👍

PR #271, opened 2024-01-05
Don't separately install drush

rosiel:
**GitHub Issue**: https://github.com/Islandora-Devops/islandora-playbook/issues/269

# What does this Pull Request do?

There are two things in this PR:
- add `vagrant_user` in the Vagrant file. Without this, the Vagrant installation doesn't work. I'm not sure how it worked before.
- fixes proposed by @RodBruce (many thanks!) for installing Drupal 10 (namely, _not_ installing the Drush Installer.)

Drush is still installed, but because it's required by the Starter Site or (hopefully) whatever you're using in place of the starter site. This is how it's supposed to be installed.

# What's new?

* Adds a value to `ansible_user` in the Vagrantfile
* Removes the `geerlingguy.drush` playbook role from `requirements.yml`
* Removes the call to that role in `webserver.yml`.

In regard to change number 3 in the linked issue, I found that i did not have to change anything in the drupal playbook for it to work. When Composer deployed the drupal project from the starter site, drush 12 was installed into the /vendor/bin folder. The [Drupal ansible role looks in this folder for Drush](https://github.com/geerlingguy/ansible-role-drupal/blob/master/tasks/main.yml#L43-L52). I ran the playbook using Vagrant with no other alterations (your change no.1 to the playbook version string has already been merged) and I got a Drupal 10.1.5 site with no Playbook errors. 

* Does this change require documentation to be updated?  No
* Does this change add any new dependencies? No
* Does this change require any other modifications to be made to the repository
 (i.e. Regeneration activity, etc.)? No
* Could this change impact execution of existing code? No

# How should this be tested?

* clone this repo
* type `vagrant up` or use the [deploy instructions](https://islandora.github.io/documentation/installation/playbook/#deploying-to-a-remote-environment) to deploy it to another server.

# Additional Notes:
Any additional information that you think would be helpful when reviewing this
 PR.

# Interested parties
Tag (@ mention) interested parties or, if unsure, @Islandora-Devops/committers

Issue #269, opened 2023-11-09
Changes needed to install Drupal 10


RodBruce:
I had to make a few changes in order to get the playbook to do a clean install of Drupal 10:

1. Update: `vars/starter.yml` to use a newer version of the Starter Site:

```
diff ~/tmp/islandora-playbook/vars/starter.yml vars/starter.yml
8c8
< drupal_composer_project_package: 'islandora/islandora-starter-site:~0.7'
---
> drupal_composer_project_package: 'islandora/islandora-starter-site:~1.2'
```
2. The way Drush gets installed and used with Drupal 10 is different. We no longer need the `geerlingguy.drush` role in `requirements.yml`

```
diff ~/tmp/islandora-playbook/requirements.yml requirements.yml
39,42d38
< 
< - src: geerlingguy.drush
<   version: 3.1.1
< 
```
and we no longer need to include it in `webserver.yml`
```
diff ~/tmp/islandora-playbook/webserver.yml webserver.yml
24,25d23
<     - name: geerlingguy.drush  # Include drush as drupal-openseadragon needs it.
<       when: islandora_build_base_box|bool == False
```
3. The `geerlingguy.drupal` role has not been updated for Drupal 10. The version we are loading (4.3.0) is the latest version on the ansible galaxy website (https://galaxy.ansible.com/ui/standalone/roles/geerlingguy/drupal/). So, after running `ansible-galaxy install` to pull in the requirements I needed to change the drush version in `roles/external/geerlingguy.drupal/defaults/main.yml` from 10.1 to 12.

```
diff ~/tmp/islandora-playbook/roles/external/geerlingguy.drupal/defaults/main.yml roles/external/geerlingguy.drupal/defaults/main.yml 
28c28
<   - "drush/drush:^10.1"
---
>   - "drush/drush:^12"
```
Would it make sense to fork that role, update it and move it to an internal role until it is updated on the galaxy site?

Thanks

whikloj:
Another option is to make a PR against, the Geerlingguy repo. https://github.com/geerlingguy/ansible-role-drupal :shrug:

BrzIGH:
Hi folks,

I just ran the ansible playbook on a new instance. It still installs Drupal 9.

Is there an ETA when the playbook will install Drupal 10 ?

Thanks a lot for the playbook....great job !

rosiel:
Ah, bit again by the tilde operator! 
[...]

PR #274, opened 2024-01-19
Unset drupal_composer_dependencies so as to not separately require drush.

rosiel:
**GitHub Issue**: #269 

Past issue that failed to account for this: #271 

Reverts a line that removes the default geerlingguy drupal role's tendency to install drush 10. We already install drush through having it in the Starter Site's composer.json so we don't need to do this.

A brief description of what the intended result of the PR will be and/or what
 problem it solves.

# What's new?

Un-breaks the playbook.

* Does this change require documentation to be updated?  no
 (i.e. Regeneration activity, etc.)?  no
* Could this change impact execution of existing code? no

# How should this be tested?

* Run the playbook now. It should fail during the "islandora_build_base_box=false" step, when doing `Install dependencies with composer require` 
* With this PR, it should go through fine.

# Additional Notes:
 PR.

# Interested parties

### DIFF
--- a/vars/starter.yml
+++ b/vars/starter.yml
@@ -2,7 +2,7 @@
 drupal_build_composer_project: true
 
 # None necessary, should all be described in the given project.
-#drupal_composer_dependencies: []
+drupal_composer_dependencies: []
 
 # Presently pointing at the main branch.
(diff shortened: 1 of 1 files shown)


seth-shaw-asu (approved):
Works for me.

Issue #276, opened 2024-04-09
Playbook failing again


BrzIGH:
Last week I tried running a playbook that has served me well since the beginning of the year. It failed.
Re-downloaded the external roles and tried again. Failed again.
Built a brand new Ubuntu 22.04.4 LTS box and redownloaded the complete playbook and external roles. It failed again.

Here is where I consistently get the error:

TASK [geerlingguy.drupal : Generate Drupal project with composer package in /tmp/composer-project (this may take a while).] *************************************************************
Tuesday 09 April 2024  14:21:59 -0700 (0:00:00.532)       0:00:11.516 ********* 
[...]

PR #277, opened 2024-05-22
Add intl extension for php

rosiel:
**GitHub Issue**: https://github.com/Islandora-Devops/islandora-playbook/issues/276

playbook fails with modern starter site since the changes to citation select that requires the citeproc library directly.

Adds the intl php extension so it works again.

# What's new?

See diff.

# How should this be tested?

* run playbook, it fails (see issue)
* with this fix, it works.

# Additional Notes:
 PR.

# Interested parties


seth-shaw-asu (approved):
:shipit:

